PR #352, opened 2017-06-23
Openstack4J integration

pawelpros:
Removed support for jcloud and updated code to use openstack4j library with keystone version 3. 

Additionally: 
- Updated floating ip behavior 
- updated deprecated usage of extensions to proper supported ones
- updated OSGI
- removed all dependencies connected with jcloud
- added dependencies to openstack4j


osc-server/src/main/java/org/osc/core/broker/rest/client/openstack/openstack4j/Openstack4JNeutron.java
```diff
@@ -298,35 +296,11 @@ public synchronized NetFloatingIP createFloatingIp(String region, String network
...
-            if (e.getStatusCode().getCode() == OPENSTACK_CONFLICT_STATUS) {
-                releaseUnusedFloatingIps(region);
-                log.info("Retrying creating floating ip to solve the problem: " + e.getMessage());
-                netFloatingIP = getOs().networking().floatingip().create(builder.build());
-            } else {
-                throw e;
-            }
-        }
-
+        NetFloatingIP netFloatingIP = getOs().networking().floatingip().create(builder.build());
```
arvindn05: rethrow the fault in case actionResponse is not a success.

for 404, you can ignore the response but log a warning

osc-server/src/main/java/org/osc/core/broker/rest/client/openstack/openstack4j/Openstack4JNeutron.java
```diff
@@ -0,0 +1,364 @@
...
+
+        Map<String, String> filter = Maps.newHashMap();
+        filter.put("status", FloatingIpStatus.DOWN.toString());
+
+        List<? extends NetFloatingIP> list = getOs().networking().floatingip().list(filter);
+        list.forEach(floatingIp -> getOs().networking().floatingip().delete(floatingIp.getId()));
+        getOs().removeRegion();
+    }
+
+    public synchronized void deleteFloatingIp(String region, String floatingIpId) {
```
arvindn05: Please address the above for line 308 as well

osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/virtualizationconnector/VCDeleteMetaTask.java
```diff
@@ -57,20 +60,27 @@ public void executeTransaction(EntityManager em) throws Exception {
...
+        } finally {
+            // Unlock VC.
             if (vcUnlockTask != null) {
                 log.info("Releasing lock for VC '" + this.vc.getName() + "'");
                 LockManager.getLockManager().releaseLock(new LockRequest(vcUnlockTask));
             }
         }
     }
 
+    private void cleanKeystoneProviderConnection(VirtualizationConnector vc) throws EncryptionException {
```
arvindn05: how does this work for jobs? is the keystone provider used for all api calls?

How does this get cleaned up in those instances?

osc-server/src/main/java/org/osc/core/broker/rest/client/openstack/openstack4j/Openstack4JNeutron.java
pawelpros: Updated code

osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/virtualizationconnector/VCDeleteMetaTask.java
pawelpros: It's solution for manipulating map as there is no overall way for clearing the map (only when application is turned down) so it basically remove used existing endpoint if it's removed from OSC or updated. In rest case i don't see place where we could perform cleaning whole map as always some jobs could use connection.

pawelpros:
I created new forked pull request

PR #354, opened 2017-06-27
SDK Dependencies updated to 1.0.1-SNAPSHOT

lakodali:


arvindn05:
the sdn version has changed...you might need to update the version for sdn

PR #363, opened 2017-07-05
remove esapi dependecy and fix #282

### LINKED ISSUE #282: Investigate Removing esapi dependency
Esapi API has a large transitive dependency graph. 

Based on the usages in the code, its only used to escape HTML. SafeHTML provided by GWT provides the equivalent functionality and possibly can be used.

OWASP/java-html-sanitizer also seems like a much more focused library for our use

scrren

arvindn05:


osc-export/server.bndrun
```diff
@@ -85,11 +85,10 @@
...
-	org.apache.commons.collections;version='[3.2.2,3.2.3)',\
-	org.apache.commons.configuration;version='[1.10.0,1.10.1)',\
-	org.apache.commons.fileupload;version='[1.3.1,1.3.2)',\
 	org.apache.commons.codec;version='[1.6.0,1.6.1)',\
 	org.apache.commons.logging;version='[1.1.3,1.1.4)',\
 	osc-server;version='[1.0.0,1.0.1)',\
 	osc-common;version='[1.0.0,1.0.1)',\
-	log4j;version='[1.2.17,1.2.18)'
\ No newline at end of file
+	log4j;version='[1.2.17,1.2.18)',\
```
emanoelxavier: Please also resolve the -debug.bdnrun. Also how are these openstack-* dependeny changes related to this overall change?

arvindn05: good catch...will update it. 

The run bundles are autogenerated as part of the resolve operation so i dont think my changes effected this...probably bndrun is stale?

emanoelxavier: likely, I saw that tim's changes had the same resolve result.

PR #370, opened 2017-07-11
Update references to tenant -> project

arvindn05:
includes changes for DB, log, UI, services,API,DTO


arvindn05:
closing pull request so api changes can be done and tested

arvindn05:
SDN controller sdk has references to tenant still. Changing that would result in a major revision bump...wanted to wait to see if we should do that. NSC plugin would have references to tenant as well since that is still using v2 keystone.

good point on the documentation...will see if that requires updates.

PR #375, opened 2017-07-14
DE3845 - When adding or editing MC, make sure the intermediate error …

dmitryintel:
…message window is shown upon socket exceptions.

If the IP address is nonexistent, or the controller a dialogue is supposed to show summarizing one of several possible things that have gone wrong (screenshot attached). At this point, the user has an option to proceed or not. The string template value for the dialogue box is the **vc.confirm.rabbit** property from the **VmidcMessage.properties**.

Decided to get the root cause exception and put its message into the string template. Screenshot is attached.

mcvcbadipdialogs

### DIFF
--- a/osc-server/src/main/java/org/osc/core/broker/service/mc/AddApplianceManagerConnectorService.java
+++ b/osc-server/src/main/java/org/osc/core/broker/service/mc/AddApplianceManagerConnectorService.java
@@ -19,7 +19,10 @@ package org.osc.core.broker.service.mc;
 import java.net.SocketException;
 import java.util.ArrayList;
 
+import javax.net.ssl.SSLException;
 import javax.persistence.EntityManager;
+import javax.ws.rs.ForbiddenException;
+import javax.ws.rs.NotAuthorizedException;
 
 import org.apache.commons.lang.exception.ExceptionUtils;
 import org.apache.log4j.Logger;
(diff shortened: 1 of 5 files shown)


dmitryintel:
Then it looks like I will have to create another vc.confirm property in VmidcMessage. properties.  We can square out the exact wording tomorrow.

dmitryintel:
*"Failed to connect to {0}"* , the {0}-parameter being " The Controller" in this case.

arvindn05:
vc has the warnings as well. they are just not being triggered correctly see https://github.com/ContainX/openstack4j/issues/1060

osc-server/src/main/java/org/osc/core/broker/service/mc/AddApplianceManagerConnectorService.java
```diff
@@ -158,7 +163,18 @@ void checkManagerConnection(DryRunRequest<ApplianceManagerConnectorRequest> requ
...
                 this.apiFactoryService.checkConnection(mc);
             } catch (Exception e) {
-                ErrorTypeException errorTypeException = new ErrorTypeException(e, ErrorType.MANAGER_CONNECTOR_EXCEPTION);
+
+                Throwable rootCause = ExceptionUtils.getRootCause(e);
+                Throwable cause = e;
+
+                if (rootCause instanceof SocketException
+                        || rootCause instanceof SSLException
+                        || rootCause instanceof ForbiddenException
```
arvindn05: looks like we cannot handle unauthorized case without adding checked exceptions to the API package. So just handle the connection exceptions case

arvindn05:
please update the branch and the pull request description based on our discussions. I will merge the request once this is done.

@dmitryintel can you create a defect/enhancement to indicate that we need checked exception definitions in the API packages to handle cases like this?
@ychoma @alin2k please update your test cases to remove the credential check.

Issue #372, opened 2017-07-12
null pointer exception when trying to upgrade distributed appliance


arvindn05:
### Expected Behavior
no null pointer and upgrade should succeed.
### Actual Behavior
getting a null pointer with the below stacktrace
```
java.lang.NullPointerException
	at org.osc.core.broker.service.tasks.conformance.openstack.deploymentspec.OsDAIUpgradeMetaTask.executeTransaction(OsDAIUpgradeMetaTask.java:107)
	at org.osc.core.broker.service.tasks.TransactionalMetaTask.lambda$execute$0(TransactionalMetaTask.java:51)
	at org.apache.aries.tx.control.service.common.impl.AbstractTransactionControlImpl$TransactionBuilderImpl.doWork(AbstractTransactionControlImpl.java:161)
	at org.apache.aries.tx.control.service.common.impl.AbstractTransactionControlImpl$TransactionBuilderImpl.required(AbstractTransactionControlImpl.java:84)
	at org.apache.aries.tx.control.service.common.impl.AbstractTransactionControlImpl.required(AbstractTransactionControlImpl.java:263)
	at org.osc.core.broker.service.tasks.TransactionalMetaTask.execute(TransactionalMetaTask.java:50)
	at org.osc.core.broker.job.TaskNode.run(TaskNode.java:308)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:748)
```

### Steps to Reproduce
1. Deploy a Distributed appliance with a service function definition of version 1
2. Allow sync job to complete
3. Edit Distributed appliance with service function definition of version 2
4. Sync job fails and the appliance on the virtualization environment is not upgraded

### Additional Information
Looks like this is related to the OSGI changes. deleteSvaServerTask is being injected as a dynamic reference, but it is being intialized to null

### Status
- [x] Reproduced
- [x] In Progress
- [ ] Validated

arvindn05:
@alin2k please verify and update the issue with the build which fixes the issue

alin2k:
[...]

PR #385, opened 2017-07-21
DE4562 -- Fix nullpointer on line 109 of OsDAIUpgradeMetaTask

dmitryintel:
Debbugging showed that this.deleteSvaServerTask was the null field responsible for the exception. Made sure it is copied over.


arvindn05:
fixes #372

PR #390, opened 2017-07-25
Fix bnd warnings

wardtimx:


arvindn05:
@wardtimx looks like the warnings on quartz are not being suppressed. I still see them in the IDE.

without your supress change i see 
`Export org.quartz,  has 3,  private references [org.quartz.spi, org.quartz.impl.matchers, org.quartz.utils]`
with your changes i see
`Export org.quartz,  has 3,  private references [org.quartz.utils, org.quartz.impl.matchers, org.quartz.spi]`

wardtimx:
@arvindn05 - I made an effort to be as specific as possible with the warning filters, rather than accidentally hiding anything else. Unfortunately this seems to be one of the places where bnd doesn't do sorting of its output, so the error message won't always be stable.

I have raised an issue against bnd to sort this better in the future (see  https://github.com/bndtools/bnd/issues/2100), but the suggested `Export org.quartz, has 3, private references *` will work fine.

PR #396, opened 2017-07-27
fix delete DS task

arvindn05:
Clear out the protected ports from the DAI before deleting the DAI.


arvindn05:
@mrajora-1 please test with 1.0.0-SNAPSHOT-33-g25f9143 version

mrajora-1:
Hi Arvind
Tested and verified with the suggested build.
Thanks
Best/Manishi

PR #399, opened 2017-07-29
DE4568 -- Fix for the bug. Workaround the possible bug in Aries entit…

dmitryintel:
…y manager.


osc-server/src/main/java/org/osc/core/broker/job/TaskNode.java
```diff
@@ -123,7 +123,10 @@ synchronized void setState(TaskState state) {
         }
 
         if (this.taskRecord != null) {
-            persistState();
+            // Maybe bug on Pessimistic lock
```
arvindn05: can you add more detailed comment? Maybe a TODO or reference to a issue filed in the OSC repo?

PR #404, opened 2017-08-01
Ensure that the UI works under live updates

wardtimx:
When running OSC inside the Bndtools IDE there can be a live update of the OSC code. This is great for rapid development cycles, but it has shown up a lifecycle problem in the OSC UI code. If the UI component is shut down without updating the UI bundle (for example by making a change to an OSC service) then invalid Vaadin sessions are left hanging.

To get this working required several small changes

* Ensure that the osc-server and osc-ui generated source is built by removing the m2e exclusion for the build-helper-maven-plugin. Without this the generated messages file is not built into the bundle, causing NoClassDefFoundError at runtime.
* Avoid NPEs when shutting down runners. The Server component did not check for null when cleaning up runners.
* Always destroy the UI on component restart. The UI servlet did not have a destroy method which forced client sessions to close
* Always tear down the server component on update. The server component did not clean up if/when it restarted

### DIFF
--- a/osc-server/src/main/java/org/osc/core/server/Server.java
+++ b/osc-server/src/main/java/org/osc/core/server/Server.java
@@ -70,6 +70,7 @@ import org.osgi.framework.ServiceRegistration;
 import org.osgi.service.component.ComponentServiceObjects;
 import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Deactivate;
 import org.osgi.service.component.annotations.Reference;
 import org.osgi.service.component.annotations.ReferenceCardinality;
 import org.osgi.service.component.annotations.ReferencePolicy;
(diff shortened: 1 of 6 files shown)


osc-server/src/main/java/org/osc/core/server/Server.java
```diff
@@ -550,8 +566,16 @@ public void shutdownRabbitMq() {
             // No problem - this means the service was
             // already unregistered (e.g. by bundle stop)
         }
-        this.deploymentSpecRunnerCSO.ungetService(this.rabbitMQRunner.getOsDeploymentSpecNotificationRunner());
-        this.securityGroupRunnerCSO.ungetService(this.rabbitMQRunner.getSecurityGroupRunner());
+        ComponentServiceObjects<OsDeploymentSpecNotificationRunner> cso = this.deploymentSpecRunnerCSO;
+        if(cso != null) {
```
arvindn05: nitpick, i would just use deploymentspecrunnerCSO instead of calling it cso and cso2

Issue #366, opened 2017-07-07
Adding multiple security group members (SGM) of the same type to the security group entity does not work


emanoelxavier:
### Expected Behavior
From within OSC code calling `SecurityGroup.addSecurityGroupMember` with members of the same type should work as expected: new and different members should be added. This issue does not seem to reflect directly on OSC APIs or UI but it does affect unit testing and it could have other not yet found side effects. 

### Actual Behavior
From within OSC code calling `SecurityGroup.addSecurityGroupMember` with members of the same type does not work. Only the first member gets added, this is happening because the type of the security members collection is `TreeSet` which uses `compareTo` to identify if elements are the same. Currently the SecurityGroupMember implements `compareTo` using the `type` only, thus adding elements of the same type does not work. This does not seem to be affecting any behavior on the APIs or UIs, the reason seems because the persistence framework instantiates the set of SGMs with a different collection other than `TreeSet` this affects the unit tests though.

### Steps to Reproduce
Since this is currently only affecting unit tests the repro steps involve removing the workaround added for the unit tests and running them:
1.  Run the unit tests AllocateDAIWithSGIMembersTaskTest 
2. Observe the unit tests are passing
3. Remove the workaround/TODO on UpdateDAIToSGIMembersTaskTest.newSGM which is currently mocking the compareTo method of the SGM
4. Rerun the unit tests AllocateDAIWithSGIMembersTaskTest and observe that they will now fail due to the TreeSet issue.

### Additional Information
Once the TreeSet issue is addressed, i.e.: use a HashSet then the tests should pass without the need for any workaround.

### Status
- [ ] Reproduced
- [ ] In Progress
- [ ] Validated

emanoelxavier:
@arvindn05 fyi

emanoelxavier:
@opensecuritycontroller/osc_triage fyi

PR #405, opened 2017-08-03
Issue 366: use Hash Set in the SecurityGroup. Also...

dmitryintel:
... remove the Mockito wrappers around the SecurityGroup in unit tests. https://github.com/opensecuritycontroller/osc-core/issues/366

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,8 +6,7 @@ osc-server-bom/root/opt/vmidc/bin/webapp/**
 osc-export/generated/**
 osc-export/webapp/**
 osc-export/log/**
-osc-export/mgr_plugins/**
-osc-export/sdn_ctrl_plugins/**
+osc-export/plugins/**
 osc-export/*.db*
 osc-export/server.pid
 osc-server-bom/root/opt/vmidc/bin/META-INF/
--- a/osc-export/pom.xml
+++ b/osc-export/pom.xml
@@ -298,13 +298,13 @@
 								<configuration>
 									<tasks>
 										<!-- copy bundled plugins for bundled profile "all". -->
-										<copy tofile="${basedir}/mgr_plugins/SmcMgrPlugin.bar"
+										<copy tofile="${basedir}/plugins/SmcMgrPlugin.bar"
 											file="${basedir}/../../security-mgr-smc-plugin/target/SmcMgrPlugin.bar" />
-										<copy tofile="${basedir}/mgr_plugins/NsmMgrPlugin.bar"
+										<copy tofile="${basedir}/plugins/NsmMgrPlugin.bar"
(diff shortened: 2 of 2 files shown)


larkinscarvalho:
Please add reviewers to the PR

osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/deploymentspec/OsDAIConformanceCheckMetaTask.java
```diff
@@ -191,6 +191,11 @@ public void executeTransaction(EntityManager em) throws Exception {
                 ApplianceSoftwareVersion currentSoftwareVersion = ds.getVirtualSystem().getApplianceSoftwareVersion();
                 boolean doesSvaVersionMatchVsVersion = false;
                 for (OsImageReference imageRef : ds.getVirtualSystem().getOsImageReference()) {
+                	
+                	if (sva.getImage() == null) {
```
arvindn05: unrelated change...please remove as its tracked by different bug

larkinscarvalho: How does this help? More details?

dmitryintel: I pushed changes relating to another bug. Inside that loop, I have seen the getImage to return null and this is the only sensible way to deal with the situation.

osc-domain/src/main/java/org/osc/core/broker/model/entities/virtualization/SecurityGroup.java
```diff
@@ -75,7 +75,7 @@
 
     @OneToMany(mappedBy = "securityGroup", fetch = FetchType.LAZY, cascade = CascadeType.ALL)
     @OrderBy("type")
-    private Set<SecurityGroupMember> securityGroupMembers = new TreeSet<>();
+    private Set<SecurityGroupMember> securityGroupMembers = new HashSet<>();
```
arvindn05: one of the reasons why we used a tree set was to have a sorted list of members. Please check to see if any code relies on this collection being sorted.

Also since we are not putting the SecurityGroupMember in a tree set....it does not need to implement comparable interface. Please update the entity as required.

emanoelxavier: For entities coming from the db observe the OrderBy above. I believe the comparable wold still make sense if there is any piece of code that is initiating this field without db data, sorting it and expecting it would be by type. I am not sure if there is such case but if not I agree that the comparable can be removed.

PR #442, opened 2017-08-10
remove NSM and SMC plugin from the all profile

arvindn05:
remove NSM and SMC from the all profile until those plugins are open sourced.
Adds nuage to the all profile since it is already part of the opensecuritycontroller org


arvindn05:
sorry for the many updates. Making some other improvements as part of this change

PR #447, opened 2017-08-22
Found an error rather than exception thrown in ServiceDispatch.exec()

dmitryintel:
The exec() code and handleException() assume a general Exception. I have seen a LinkageError thrown once.  

That resulted in the class cast exception in the following calls:
`   handleException((Exception) e.getCause());`


larkinscarvalho:
@dmitryintel Please fix the unit tests

PR #453, opened 2017-08-31
Adding the method in the k8s wrapper package to check the service rea…

emanoelxavier:
…diness


osc-server/src/test/java/org/osc/core/broker/rest/client/k8s/KubernetesStatusApiTest.java
```diff
@@ -0,0 +1,167 @@
...
+
+        // Act.
+        boolean isServiceReady = this.api.isServiceReady();
+
+        // Assert.
+        assertFalse(isServiceReady);
+    }
+
+    @Test
+    public void testIsServiceReady_WhenComponentStatusConditionFalse_ServiceReady() throws Exception {
```
karimull: isn't it should be testIsServiceReady_WhenComponentStatusConditionTrue_ServiceReady

emanoelxavier: @karimull i will update the test name above, yes it should be true. For the other comment... we dont need a unit test specific to validate if the component-name is null or wrong because if that is regressed in the code the mocked line when(this.mixedOperationMock.withName("controller-manager")).thenReturn(componentStatusResource); will not work causing many other tests to fail therefore catching the regression. In other words, that mock line is already enforcing that.

osc-server/src/test/java/org/osc/core/broker/rest/client/k8s/KubernetesStatusApiTest.java
```diff
@@ -0,0 +1,167 @@
...
+        boolean isServiceReady = this.api.isServiceReady();
+
+        // Assert.
+        assertFalse(isServiceReady);
+    }
+
+    @Test
+    public void testIsServiceReady_WhenComponentStatusConditionTrue_ServiceReady() throws Exception {
+        // Arrange.
+        ComponentStatus componentStatus = newComponentStatus("Healthy", "True");
```
arvindn05: do you want to reference the constants defined in the KubernetesStatusApi? You can make them package private for this

osc-server/src/test/java/org/osc/core/broker/rest/client/k8s/KubernetesStatusApiTest.java
```diff
@@ -0,0 +1,167 @@
...
+        when(componentStatusResource.get()).thenReturn(result);
+        when(this.mixedOperationMock.withName("controller-manager")).thenReturn(componentStatusResource);
+    }
+
+    private ComponentStatus newComponentStatus(String conditionType, String conditionStatus) {
+        ComponentStatus componentStatus = new ComponentStatus();
+        List<ComponentCondition> conditions = new ArrayList<ComponentCondition>();
+
+        if (conditionType != null) {
+            ComponentCondition condition = new ComponentCondition();
```
arvindn05: can be simplified into componentStatus.setConditions(Arrays.asList(new ComponentCondition(null, null, conditionType, conditionStatus)

osc-server/src/test/java/org/osc/core/broker/rest/client/k8s/KubernetesStatusApiTest.java
emanoelxavier: I wanted to actually use the strings. If the same string fields are referenced and they get changed accidentally (regression scenario) the tests wont flag that since the value is shared. If someone is intentionally changing the strings they must update the tests as well and the failure serves as a "are you sure you want to change this?" type of warning.

arvindn05: makes sense. although unit tests should be testing the logic not the state though...but its fine. If these values change on k8s we will have to update in all the places...since its only 2 its less of a problem.

PR #455, opened 2017-09-01
Enabling k8s virtualization connector

emanoelxavier:
1. Adding KUBERNETES VC type
2.  Replacing `virtualizationConnectorUtil.checkOpenstackConnection` with `checkConnection` . `checkConnection` uses `checkOpenstackConnection` and the **newly** added checkK8sConnection. Both of them use the same `checkSDNControllerConnection`. 
3. Avoiding sharing test data across different test classes and packages
Previously it seems that the `AddVCService` was refactored into smaller classes: validators and check connection util but their test data was still referencing a single test data file. This made all these tests a lot harder to update, read and maintain: i.e.: when the data is changed one needs to know which tests are using it and update them as well. This change isolates each test class with their own test data keeping them package private. This is consistent with the approach adopted for other test classes within the code base as well: class file, test class file, test data file. 
3. Additional unit tests:
* Enforcing that K8s VC must have an SDN defined
* Unit tests to ensure K8s connectivity is checked during VC validation
5. Small UI changes to allow K8s VC to be added

### DIFF
--- a/osc-server/src/main/java/org/osc/core/broker/util/VirtualizationConnectorUtil.java
+++ b/osc-server/src/main/java/org/osc/core/broker/util/VirtualizationConnectorUtil.java
@@ -16,15 +16,22 @@
  *******************************************************************************/
 package org.osc.core.broker.util;
 
-import com.rabbitmq.client.ShutdownSignalException;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
+
 import org.apache.log4j.Logger;
 import org.osc.core.broker.model.entities.virtualization.VirtualizationConnector;
 import org.osc.core.broker.model.plugin.ApiFactoryService;
+import org.osc.core.broker.rest.client.k8s.KubernetesClient;
+import org.osc.core.broker.rest.client.k8s.KubernetesStatusApi;
 import org.osc.core.broker.rest.client.openstack.openstack4j.Endpoint;
 import org.osc.core.broker.rest.client.openstack.openstack4j.Openstack4jKeystone;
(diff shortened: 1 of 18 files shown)


osc-server/src/main/java/org/osc/core/broker/util/VirtualizationConnectorUtil.java
```diff
@@ -75,82 +83,19 @@ private void delayedInit() {
...
-                    LOG.warn(
-                            "Exception encountered when trying to connect to RabbitMQ, allowing user to either ignore or correct issue");
-                }
+            if (vc.getVirtualizationType().equals(VirtualizationType.OPENSTACK)) {
+                errorTypeException = checkOpenstackConnection(request, certificateResolverModels, vc);
+            } else if (vc.getVirtualizationType().equals(VirtualizationType.KUBERNETES)) {
+                errorTypeException = checkKubernetesConnection(request, certificateResolverModels, vc);
             }
 
             this.managerFactory.clearListener();
```
arvindn05: seems odd place for this to be....move it to where the manager factory listener is added

emanoelxavier: Can you be more specific, i.e.: which ln? This has not changed but I can move if there is no risk of regression in doing so.  As it was before there are multiple places where the listener is being initiated, all of them private methods, this being the common public method.

arvindn05: i see initSSLCertificatesListener(this.managerFactory, certificateResolverModels, "rabbitmq"); method calls and the clear listener is the counterpart for it. With this refactor, i dont see where the listeners are initialized. It makes sense to keep them together(atleast in the same method) so there is clarity

PR #456, opened 2017-09-01
 SFC database schema table and entity creation

karimull:
Created a sfc-entity to Virtual system many to many relation ship


osc-domain/src/main/java/org/osc/core/broker/model/entities/virtualization/ServiceFunctionChain.java
```diff
@@ -0,0 +1,74 @@
...
+@Entity
+@Table(name = "SERVICE_FUNCTION_CHAIN", uniqueConstraints = { @UniqueConstraint(columnNames = { "name" }) })
+public class ServiceFunctionChain extends BaseEntity {
+
+	@Column(name = "name", nullable = false)
+	private String name;
+
+	@ManyToMany(fetch = FetchType.LAZY)
+	@JoinTable(name = "SERVICE_FUNCTION_CHAIN_VIRTUAL_SYSTEM", joinColumns = @JoinColumn(name = "sfc_fk", referencedColumnName = "id"), inverseJoinColumns = @JoinColumn(name = "virtual_system_fk", referencedColumnName = "id"))
+	private Set<VirtualSystem> sfcVs = new HashSet<>();
```
arvindn05: should just be virtualSystems

larkinscarvalho: Refactor to virtualSystems. Update mapping.

karimull:
Addressed comments on Attribute name change

PR #458, opened 2017-09-06
Adding pod, podport and label entities and schema updates

emanoelxavier:


emanoelxavier:
Looks like this might be failing on Travis for some unrelated reason. I am restarting the build, the same code built on Jenkins.

osc-domain/src/main/java/org/osc/core/broker/model/entities/virtualization/k8s/Pod.java
```diff
@@ -0,0 +1,52 @@
...
+    @Column(name = "name", nullable = false)
+    private String name;
+
+    @Column(name = "namespace", nullable = false)
+    private String namespace;
+
+    @Column(name = "external_id", nullable = false, unique = true)
+    private String externalId;
+
+    @Column(name = "node")
```
dmitryintel: Are the names not the same by default?

osc-domain/src/main/java/org/osc/core/broker/model/entities/virtualization/k8s/PodPort.java
```diff
@@ -0,0 +1,56 @@
...
+@Entity
+@Table(name = "POD_PORT")
+public class PodPort extends BaseEntity {
+    @Column(name = "external_id", nullable = false, unique = true)
+    private String externalId;
+
+    @Column(name = "mac_address", nullable = false, unique = true)
+    private String macAddress;
+
+    @ElementCollection(fetch = FetchType.LAZY)
```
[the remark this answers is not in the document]
emanoelxavier: Likely yes however for most of the cases when this entity is being loaded the ip addresses will be used. Also this is consistent with other similar entities, like VMPort

osc-domain/src/main/java/org/osc/core/broker/model/entities/virtualization/k8s/Pod.java
emanoelxavier: Likely yes but I think being explicit is better since in some cases it is not a perfect match.

emanoelxavier:
@dmitryintel Please let me know if you have any further questions/comments or if it can be approved.

PR #463, opened 2017-09-07
Enabling creation of security group and members using k8s labels

emanoelxavier:
This PR is built on top of #458  when reviewing it look at the last commit only, the first two commits will be rebased after #458 merge.

This PR allows the creation of SGs and SG members for K8s of type label. Synchronizing K8s SGs is out of scope for this PR and it will be handled later.


osc-domain/src/main/java/org/osc/core/broker/model/entities/virtualization/SecurityGroupMember.java
```diff
@@ -69,11 +69,13 @@
     @Column(name = "address")
     private String address;
 
-    public SecurityGroupMember(SecurityGroup securityGroup, OsProtectionEntity entity) {
+    public SecurityGroupMember(SecurityGroup securityGroup, ProtectionEntity entity) {
         this.securityGroup = securityGroup;
         this.type = entity.getType();
         if (this.type == SecurityGroupMemberType.VM) {
             this.vm = (VM) entity;
+        } else if (this.type == SecurityGroupMemberType.LABEL) {
```
dmitryintel: import static?

osc-domain/src/main/java/org/osc/core/broker/model/entities/virtualization/k8s/Label.java
```diff
@@ -29,15 +29,23 @@
...
 import org.osc.core.broker.model.entities.virtualization.SecurityGroupMember;
+import org.osc.core.broker.model.entities.virtualization.SecurityGroupMemberType;
 
 @SuppressWarnings("serial")
 @Entity
 @Table(name = "LABEL")
-public class Label extends BaseEntity {
-    @Column(name = "value", nullable = false)
+public class Label extends BaseEntity implements ProtectionEntity {
+    public Label() {
```
arvindn05: move constructor after variables and make it package private or private

osc-domain/src/main/java/org/osc/core/broker/model/entities/virtualization/k8s/Label.java
```diff
@@ -29,15 +29,23 @@
...
+    }
+
+    @Column(name = "value", nullable = false, unique = true)
```
arvindn05: name and value may not be unique. The label/port and other entities are shared across vc's

osc-server/src/main/java/org/osc/core/broker/service/persistence/LabelEntityMgr.java
```diff
@@ -0,0 +1,45 @@
...
+import javax.persistence.EntityManager;
+import javax.persistence.NoResultException;
+import javax.persistence.criteria.CriteriaBuilder;
+import javax.persistence.criteria.CriteriaQuery;
+import javax.persistence.criteria.Root;
+
+import org.osc.core.broker.model.entities.virtualization.k8s.Label;
+
+public class LabelEntityMgr {
+    public static Label findByValue(EntityManager em, String labelValue) {
```
arvindn05: might need to pass vc

osc-server/src/main/java/org/osc/core/broker/service/securitygroup/AddSecurityGroupService.java
```diff
@@ -85,27 +85,36 @@ public BaseJobResponse exec(AddOrUpdateSecurityGroupRequest request, EntityManag
...
-                    LockUtil.releaseLocks(forLambda);
-                    throw e;
-                }
-            });
-
-            // Lock the SG with a write lock and allow it to be unlocked at the end of the job
-            unlockTask.addUnlockTask(LockUtil.tryLockSecurityGroupOnly(securityGroup));
+            // TODO emanoel: remove this condition when the SG sync is implemented for K8s.
+            if (vc.getVirtualizationType().isOpenstack()) {
+                unlockTask = LockUtil.tryLockVC(vc, LockType.READ_LOCK);
```
arvindn05: we should read lock VC before we even add security group members

osc-server/src/main/java/org/osc/core/broker/service/securitygroup/UpdateSecurityGroupService.java
```diff
@@ -97,17 +102,23 @@ public BaseJobResponse exec(AddOrUpdateSecurityGroupRequest request, EntityManag
...
-
-                    return new BaseJobResponse(securityGroup.getId(), job.getId());
-                } catch (Exception e) {
-                    LockUtil.releaseLocks(forLambda);
-                    throw e;
-                }
-            });
+            // TODO emanoel: remove this condition once sync is implemented for k8s SGs.
+            if (vc.getVirtualizationType().isOpenstack()) {
+                unlockTask = LockUtil.tryLockSecurityGroup(securityGroup,
```
arvindn05: same comment as before. we need to lock much earlier

osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/securitygroup/ForceDeleteSecurityGroupTask.java
```diff
@@ -107,6 +109,20 @@ public void executeTransaction(EntityManager em) {
...
+                for (Pod pod : sgm.getLabel().getPods()) {
+
+                    for (PodPort port : pod.getPorts()) {
+                        OSCEntityManager.delete(em, port, this.txBroadcastUtil);
+                    }
+
+                    OSCEntityManager.delete(em, pod, this.txBroadcastUtil);
+                }
+
+                // remove Network from database
```
arvindn05: remove label

osc-server/src/main/java/org/osc/core/broker/util/db/upgrade/ReleaseUpgradeMgr.java
```diff
@@ -323,6 +323,7 @@ private static void upgrade84to85(Statement stmt) throws SQLException {
                 "updated_by varchar(255)," +
                 "updated_timestamp timestamp," +
                 "version bigint," +
```
arvindn05: we have to create a new DB step unless you know no one is at this db version

osc-server/src/main/java/org/osc/core/broker/service/validator/SecurityGroupDtoValidator.java
```diff
@@ -43,10 +43,6 @@ public static void checkForNullFields(SecurityGroupDto dto) throws Exception {
         Map<String, Object> map = new HashMap<String, Object>();
 
         map.put("Name", dto.getName());
```
arvindn05: where are these validated now?

osc-domain/src/main/java/org/osc/core/broker/model/entities/virtualization/k8s/Label.java
emanoelxavier: the persistence frameworks requires it to be public.

arvindn05: package private should work. it just needs a default constuctor...

osc-server/src/main/java/org/osc/core/broker/service/validator/SecurityGroupDtoValidator.java
emanoelxavier: On BaseSecGroupService ln 99

osc-domain/src/main/java/org/osc/core/broker/model/entities/virtualization/k8s/Label.java
```diff
@@ -37,13 +37,10 @@
...
 public class Label extends BaseEntity implements ProtectionEntity {
-    public Label() {
-    }
-
-    @Column(name = "value", nullable = false, unique = true)
+    @Column(name = "value", nullable = false)
     private String value;
 
-    @Column(name = "name", nullable = false, unique = true)
+    @Column(name = "name", nullable = false)
```
arvindn05: i would have expected db upgrade step to change based on these fields being unique...for future make sure the annotation and the db matches.

PR #466, opened 2017-09-11
Adding API for creating an appliance for a container VNF deployment

emanoelxavier:


osc-server/src/main/java/org/osc/core/broker/service/appliance/AddApplianceService.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @Override
+    public BaseResponse exec(BaseRequest<ApplianceDto> request, EntityManager em) throws Exception {
+        BaseResponse response = new BaseResponse();
+
+        ApplianceDto applianceDto = request.getDto();
+
+        this.applianceDtoValidator.validateForCreate(applianceDto);
+
+        Appliance appliance = new Appliance();
```
arvindn05: use parameterized constructor

osc-service-api/src/main/java/org/osc/core/broker/service/dto/ApplianceDto.java
```diff
@@ -29,6 +29,15 @@
     private String managerType;
     private String managerVersion;
 
+    public ApplianceDto() {
```
arvindn05: package private

osc-rest-server/src/main/java/org/osc/core/broker/rest/server/api/ApplianceApis.java
```diff
@@ -115,11 +119,26 @@
...
+        .getListResponse(this.listApplianceService, new BaseRequest<BaseDto>(true));
 
         return response.getList();
     }
 
+    @ApiOperation(value = "Creates a new appliance for a software function model",
+            notes = "Creates a new appliance for a software function model",
+            response = BaseResponse.class)
+    @ApiResponses(value = { @ApiResponse(code = 200, message = "Successful operation"),
+            @ApiResponse(code = 400, message = "In case of any error", response = ErrorCodeDto.class) })
```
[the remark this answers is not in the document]
emanoelxavier: Ideally we should not return 500 in any situation. That usually means something is wrong on the server but it is possible. We should clean this documentation up with the various possible status code, for the different situations.

osc-rest-server/src/main/java/org/osc/core/broker/rest/server/api/ApplianceApis.java
```diff
@@ -115,11 +119,26 @@
 
         @SuppressWarnings("unchecked")
         ListResponse<ApplianceDto> response = (ListResponse<ApplianceDto>) this.apiUtil
-                .getListResponse(this.listApplianceService, new BaseRequest<BaseDto>(true));
+        .getListResponse(this.listApplianceService, new BaseRequest<BaseDto>(true));
```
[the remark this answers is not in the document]
emanoelxavier: I have set the eclipse save actions to automatically indent the code, this is just following the settings.

osc-rest-server/src/main/java/org/osc/core/broker/rest/server/api/ApplianceApis.java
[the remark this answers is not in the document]
arvindn05: the response are not intended to ennumerate all the possible status code(like 404 etc), just one which OSC throws itself.

Everything else is based on REST specification.

osc-domain/src/main/java/org/osc/core/broker/model/entities/virtualization/ServiceFunctionChain.java
```diff
@@ -36,34 +36,37 @@
...
-	@JoinTable(name = "SERVICE_FUNCTION_CHAIN_VIRTUAL_SYSTEM", joinColumns = @JoinColumn(name = "sfc_fk", referencedColumnName = "id"), inverseJoinColumns = @JoinColumn(name = "virtual_system_fk", referencedColumnName = "id"))
-	private Set<VirtualSystem> virtualSystems = new HashSet<>();
+    @ManyToMany(fetch = FetchType.LAZY)
+    @JoinTable(name = "SERVICE_FUNCTION_CHAIN_VIRTUAL_SYSTEM", joinColumns = @JoinColumn(name = "sfc_fk", referencedColumnName = "id"), inverseJoinColumns = @JoinColumn(name = "virtual_system_fk", referencedColumnName = "id"))
+    private Set<VirtualSystem> virtualSystems = new HashSet<>();
 
-	public ServiceFunctionChain(String name) {
-		this.name = name;
-	}
+    public ServiceFunctionChain(String name) {
```
arvindn05: least to most complicated constructor.Move package private constructor below variable declarations

osc-server/src/main/java/org/osc/core/broker/service/GetDtoFromEntityService.java
```diff
@@ -132,7 +132,7 @@
             res.setDto((R) dto);
         } else if (entityName.equals("Appliance")) {
             Appliance entity = getEntity(entityId, entityName, Appliance.class, em);
-            ApplianceDto dto = new ApplianceDto();
+            ApplianceDto dto = new ApplianceDto(null, null, null);
```
arvindn05: passing null, null, null in constructor is probably a bad idea. Since this is a valid case where dto is initialized to null...lets just use the public contructor of ApplianceDto

PR #468, opened 2017-09-13
Adding POST for ASV and all the db entity changes for container VNF deployment

emanoelxavier:
This PR implements:
1. All the code needed to create an appliance software version (entity changes, services, unit tests, rest API). see https://github.com/opensecuritycontroller/community/blob/master/designs/containers/vnf-deployment/vnf-deployment.md#appliance-software-version 

2. All the DB entity changes related to DS and DAIS for K8s, see https://github.com/opensecuritycontroller/community/blob/master/designs/containers/vnf-deployment/vnf-deployment.md#osc-entities 

3.  Makes the ASV integer field using `Integer` rather than `int`. Those fields were already optional but when not provided we had an NPE. This fixes that problem.

Note: Remaining changes to create a k8s DS will come later in a different PR

### DIFF
--- a/osc-domain/src/main/java/org/osc/core/broker/model/entities/appliance/DistributedApplianceInstance.java
+++ b/osc-domain/src/main/java/org/osc/core/broker/model/entities/appliance/DistributedApplianceInstance.java
@@ -167,6 +167,30 @@ public class DistributedApplianceInstance extends BaseEntity {
         this.ipAddress = ipAddress;
     }
 
+    public String getExternalId() {
+        return this.externalId;
+    }
+
+    public void setExternalId(String externalId) {
+        this.externalId = externalId;
+    }
+
+    public String getInspectionElementId() {
+        return this.inspectionElementId;
+    }
+
+    public void setInspectionElementId(String inspectionElementId) {
+        this.inspectionElementId = inspectionElementId;
+    }
+
+    public String getInspectionElementParentId() {
+        return this.inspectionElementParentId;
+    }
+
+    public void setInspectionElementParentId(String inspectionElementParentId) {
+        this.inspectionElementParentId = inspectionElementParentId;
+    }
+
     public Date getLastStatus() {
         return this.lastStatus;
     }
--- a/osc-domain/src/main/java/org/osc/core/broker/model/entities/appliance/DistributedApplianceInstance.java
+++ b/osc-domain/src/main/java/org/osc/core/broker/model/entities/appliance/DistributedApplianceInstance.java
@@ -48,12 +48,12 @@ public class DistributedApplianceInstance extends BaseEntity {
 
     @ManyToOne(fetch = FetchType.EAGER)
     @JoinColumn(name = "virtual_system_fk", nullable = false,
-            foreignKey = @ForeignKey(name = "FK_DAI_VIRTUAL_SYSTEM"))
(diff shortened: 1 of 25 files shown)


osc-server/src/main/java/org/osc/core/broker/util/db/upgrade/ReleaseUpgradeMgr.java
```diff
@@ -239,6 +239,8 @@ private static void performUpdateChain(int curDbVer, Statement stmt, EncryptionA
                 upgrade84to85(stmt);
             case 85:
                 upgrade85to86(stmt);
+            case 86:
+                upgrade86to87(stmt);
```
dmitryintel: I think we are clashing here. I also have 87.

osc-server/src/main/java/org/osc/core/broker/util/db/upgrade/ReleaseUpgradeMgr.java
```diff
@@ -250,6 +252,31 @@ private static void performUpdateChain(int curDbVer, Statement stmt, EncryptionA
         }
     }
     
+    private static void upgrade86to87(Statement stmt) throws SQLException {
+        execSql(stmt, "alter table APPLIANCE_SOFTWARE_VERSION add column image_pull_secret_name varchar(255);");
```
dmitryintel: I have also seen some changes to the Virtual System

osc-domain/src/main/java/org/osc/core/broker/model/entities/appliance/DistributedApplianceInstance.java
```diff
@@ -48,12 +48,12 @@
 
     @ManyToOne(fetch = FetchType.EAGER)
     @JoinColumn(name = "virtual_system_fk", nullable = false,
-            foreignKey = @ForeignKey(name = "FK_DAI_VIRTUAL_SYSTEM"))
+    foreignKey = @ForeignKey(name = "FK_DAI_VIRTUAL_SYSTEM"))
```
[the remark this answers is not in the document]
emanoelxavier: I have my eclipse settings to auto-indent when I save. This is just following the settings which we should have in common I believe.

osc-server/src/main/java/org/osc/core/broker/util/db/upgrade/ReleaseUpgradeMgr.java
emanoelxavier: Whoever does the merge after needs to handle the conflict as usual.

osc-domain/src/main/java/org/osc/core/broker/model/entities/appliance/DistributedApplianceInstance.java
```diff
@@ -82,6 +82,13 @@
     @Column(name = "os_server_id")
     private String osServerId;
 
+    @Column(name = "external_id")
+    private String externalId;
+    @Column(name = "inspection_element_id")
+    private String inspectionElementId;
+    @Column(name = "inspection_element_parent_id")
+    private String inspectionElementParentId;
```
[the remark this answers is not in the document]
emanoelxavier: addressed

emanoelxavier:
The latest commit on this PR adds the remaining code to allow for a creation of a K8s deployment spec. See https://github.com/opensecuritycontroller/community/blob/master/designs/containers/vnf-deployment/vnf-deployment.md#deployment-spec

PR #470, opened 2017-09-15
Fixing response payload for creation and updates of deployment specs for K8s

emanoelxavier:


osc-server/src/main/java/org/osc/core/broker/service/AddDeploymentSpecService.java
```diff
@@ -93,6 +93,7 @@ public BaseJobResponse exec(BaseRequest<DeploymentSpecDto> request, EntityManage
                 BaseJobResponse response = new BaseJobResponse();
                 response.setId(ds.getId());
                 LockUtil.releaseLocks(unlockTask);
+                return response;
```
arvindn05: on line 86 to be consistent return new BaseJobResponse(this.ds.getId(), job.getId())

PR #474, opened 2017-09-19
allow floating ip for non-admin tenants

arvindn05:
Floating ip pool/external networks are not tenant specific. Floating pools/external networks are typically created by admin tenant. We should not filter by tenant id.

### DIFF
--- a/osc-server/src/main/java/org/osc/core/broker/rest/client/openstack/openstack4j/Openstack4JNeutron.java
+++ b/osc-server/src/main/java/org/osc/core/broker/rest/client/openstack/openstack4j/Openstack4JNeutron.java
@@ -50,7 +50,6 @@ public class Openstack4JNeutron extends BaseOpenstack4jApi {
 
     private static final String QUERY_PARAM_COMPUTE_DEVICE_OWNER = "compute:";
     private static final String QUERY_PARAM_ROUTER_DEVICE_OWNER = "network:router_interface";
-    private static final String QUERY_PARAM_TENANT_ID = "tenant_id";
     private static final String QUERY_PARAM_EXTERNAL_ROUTER = "router:external";
 
     private static final int OPENSTACK_CONFLICT_STATUS = 409;
@@ -66,7 +65,7 @@ public class Openstack4JNeutron extends BaseOpenstack4jApi {
     public List<Network> listNetworkByProject(String region, String projectId) {
         getOs().useRegion(region);
         List<? extends Network> list = getOs().networking().network().list().stream()
-                .filter(subnet -> (projectId.equals(subnet.getTenantId()) && !subnet.isShared())
+                .filter(subnet -> projectId.equals(subnet.getTenantId()) && !subnet.isShared()
                         || subnet.isShared()).collect(Collectors.toList());
         return ImmutableList.<Network>builder().addAll(list).build();
     }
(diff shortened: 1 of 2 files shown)


osc-server/src/main/java/org/osc/core/broker/rest/client/openstack/openstack4j/Openstack4JNeutron.java
```diff
@@ -50,7 +50,6 @@
 
     private static final String QUERY_PARAM_COMPUTE_DEVICE_OWNER = "compute:";
     private static final String QUERY_PARAM_ROUTER_DEVICE_OWNER = "network:router_interface";
-    private static final String QUERY_PARAM_TENANT_ID = "tenant_id";
     private static final String QUERY_PARAM_EXTERNAL_ROUTER = "router:external";
```
karimull: Is router: "external"  in related to our ext-net type "external" or opnestack4j representation of ext-net?. 
because ext-net type can be default sometimes.

osc-server/src/main/java/org/osc/core/broker/rest/client/openstack/openstack4j/Openstack4JNeutron.java
```diff
@@ -156,7 +155,7 @@ public String getNetworkPortRouterDeviceId(String projectId, String region, Port
     private Port getPortByMacAddress(String region, String macAddress) {
         getOs().useRegion(region);
         List<? extends Port> portList = getOs().networking().port().list(PortListOptions.create().macAddress(macAddress));
-        return (portList == null || portList.isEmpty()) ? null : portList.get(0);
+        return portList == null || portList.isEmpty() ? null : portList.get(0);
```
karimull: usage of CollectionUtils.emptyIfNull  might make is compact?.

osc-server/src/main/java/org/osc/core/broker/rest/client/openstack/openstack4j/Openstack4JNeutron.java
```diff
@@ -258,11 +257,10 @@ private String getVmIdByPort(Port port) {
     }
 
     // Floating IP API
-    public List<String> getFloatingIpPools(String region, String projectId) throws Exception {
+    public List<String> getFloatingIpPools(String region) throws Exception {
         getOs().useRegion(region);
 
         Map<String, String> filter = Maps.newHashMap();
-        filter.put(QUERY_PARAM_TENANT_ID, projectId);
         filter.put(QUERY_PARAM_EXTERNAL_ROUTER, Boolean.TRUE.toString());
```
karimull: As mentioned in first commet.. looks like route:external is a configuration map.

osc-server/src/main/java/org/osc/core/broker/rest/client/openstack/openstack4j/Openstack4JNeutron.java
arvindn05: its the openstack representation

osc-server/src/main/java/org/osc/core/broker/rest/client/openstack/openstack4j/Openstack4JNeutron.java
arvindn05: we can replace by CollectionUtils.emptyIfNull(portList).isEmpty() but not sure it adds lot of value

PR #476, opened 2017-09-19
Adding tasks and metatasks for K8s deployment spec conformance

emanoelxavier:
This PR includes the tasks and metatasks for conformance of k8s deployment specs and the new k8s wrapper package APIs for Deployments. See         https://github.com/opensecuritycontroller/community/blob/master/designs/containers/vnf-deployment/vnf-deployment.md#osc-synchronization-tasks


osc-server/src/main/java/org/osc/core/broker/rest/client/k8s/KubernetesDeploymentApi.java
```diff
@@ -0,0 +1,210 @@
...
+
+        if (name == null) {
+            throw new IllegalArgumentException("Name should not be null");
+        }
+
+        if (namespace == null) {
+            throw new IllegalArgumentException("Namespace should not be null");
+        }
+
+        KubernetesDeployment deployment = getDeploymentByName(namespace, name);
```
dmitryintel: So, namespace and name uniquely identify it, but we also need to know the id?

osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/k8s/deploymentspec/ConformK8sDeploymentPodsMetaTask.java
```diff
@@ -0,0 +1,66 @@
...
+    public ConformK8sDeploymentPodsMetaTask create(DeploymentSpec ds) {
+        ConformK8sDeploymentPodsMetaTask task = new ConformK8sDeploymentPodsMetaTask();
+        task.ds = ds;
+
+        return task;
+    }
+
+    @Override
+    public void executeTransaction(EntityManager em) throws Exception {
+        this.tg = new TaskGraph();
```
dmitryintel: Unimplemented, I presume

osc-server/src/main/java/org/osc/core/broker/rest/client/k8s/KubernetesDeploymentApi.java
emanoelxavier: Correct, the name and namespace are unique in space, not in time. In other words, if an entity is deleted from k8s with a given name and namespace another can be recreated with the same namespace/name however the id will be different. We want to make sure we are referring to the correct entity also in time.

osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/k8s/deploymentspec/ConformK8sDeploymentPodsMetaTask.java
emanoelxavier: Yes, it will be implemented by a future, soon to come, PR. I dont want to throw an exception at the moment because I want to see if the jobs pass right now. This task is added to a parent graph so if i throw it will fail the graph.

osc-server/src/test/java/org/osc/core/broker/service/tasks/conformance/k8s/deploymentspec/UpdateK8sDeploymentTaskTest.java
```diff
@@ -0,0 +1,112 @@
...
+    public void testExecute_WhenUpdatingSucceeds_K8sDeploymentIsUpdatedWithNewInstanceCount() throws Exception {
+        // Arrange.
+        VirtualizationConnector vc = new VirtualizationConnector();
+        vc.setProviderIpAddress("1.1.1.1");
+        VirtualSystem vs = new VirtualSystem(null);
+        vs.setVirtualizationConnector(vc);
+        vs.setId(102L);
+
+        ApplianceSoftwareVersion avs = new ApplianceSoftwareVersion();
+        avs.setImageUrl("ds-image-url");
```
dmitryintel: asv

PR #478, opened 2017-09-19
Integrate with neutron SFC Plugin

arvindn05:
Integrate with neutron SFC plugin.

Updated API's to use the "Inspection port grouping" functionality provided by neutron sfc.
Note that SFC does not support "Inspected port grouping" which is the property exposed by the SDN controller API plugin SUPPORT_PORT_GROUP

### DIFF
--- a/osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/deploymentspec/OsSvaServerCreateTask.java
+++ b/osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/deploymentspec/OsSvaServerCreateTask.java
@@ -116,12 +116,8 @@ public class OsSvaServerCreateTask extends TransactionalTask {
 
         VirtualizationConnector vc = vs.getVirtualizationConnector();
         Endpoint endPoint = new Endpoint(vc, ds.getProjectName());
-        SdnRedirectionApi controller = null;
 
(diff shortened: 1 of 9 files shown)


osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/deploymentspec/OsSvaServerCreateTask.java
```diff
@@ -160,41 +156,44 @@ public void executeTransaction(EntityManager em) throws Exception {
...
-                        portGroupId, this.dai.getName(), this.dai.getId(), ds.getName(), ds.getId()));
-
-                if (!pgAlreadyCreatedByOther) {
-                    ds = em.find(DeploymentSpec.class, ds.getId());
-                    ds.setPortGroupId(portGroupId);
-                    OSCEntityManager.update(em, ds, this.txBroadcastUtil);
-                }
-            } finally {
-                controller.close();
+                controller.registerInspectionPort(
```
dmitryintel: This happens whether or not the call on line 170 happens.

arvindn05: addressed

Issue #347, opened 2017-06-14
Inspection ports are deleted in the SDN controller only if it supports port group


emanoelxavier:
Currently the task DeleteInspectionPortTask task is only being executed if the SDN controller supports port group. This task should be included for ALL sdn controllers.

arvindn05:
Should be fixed as part of the SFC feature.

arvindn05:
@dmitryintel DeleteSvaServerAndDAIMetaTask has this check

arvindn05:
@mrajora-1 please test after dmitrys changes and close

mrajora-1:
Tested and verified earlier right after the fix but do not remember the build. Verified with NSC and SFC.

PR #483, opened 2017-09-20
Launch the inspection port removal task unconditionally.

dmitryintel:
Intends to fix issue 347; 
https://github.com/opensecuritycontroller/osc-core/issues/347


osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/deploymentspec/OsDAIConformanceCheckMetaTask.java
```diff
@@ -66,6 +66,9 @@
     DeleteSvaServerTask deleteSvaServerTask;
 
     @Reference
+    DeleteInspectionPortTask deleteInspectionPortTask;
```
arvindn05: you might get an NPE for this task as its not part of the create constructor on line 132

PR #485, opened 2017-09-20
Fix Openstack api calls

arvindn05:
fix issue where unsuccessful plugin uploads are not being deleted in import plugin service
fix issue where stale management port is being used to allocate floating ip resulting in error
Fix issue in create server call where the management port is not the first port attached to the server

Update the openstack4j facades to log for null/not found, throw exceptions/handle in case expected not null values are null

Where ever null is expected state, not adding explicit log messages and making sure the calling code handles the null state.

For List API's always adding the log messages in case of empty results.


osc-server/src/main/java/org/osc/core/broker/rest/client/openstack/openstack4j/Openstack4JNeutron.java
```diff
@@ -149,13 +164,21 @@ public String getNetworkPortRouterDeviceId(String projectId, String region, Port
 
     private List<? extends Port> listPorts(String region, String projectId, String networkId) {
         getOs().useRegion(region);
-        return getOs().networking().port().list(PortListOptions.create().networkId(networkId).tenantId(projectId));
+        List<? extends Port> portList = getOs().networking().port().list(PortListOptions.create().networkId(networkId).tenantId(projectId));
+        if (CollectionUtils.isEmpty(portList)) {
+            log.info(String.format("No ports found in network: %s for project: %s in region: %s", networkId, projectId,
```
emanoelxavier: Should we return `CollectionUtils.emptyIfNull(x)` for everything that returns a collection? Since this is our library/package we can guarantee that we never return null which facilitates for callers not needing to do null checks. This design expectation/restriction also simplifies the unit testing of the callers (each condition/check for every place this is called adds to a unit test). So instead we can add the expectation enforcement here with a single unit test for this method that fails if this method returns null when its dependency/mocked call returns null.

arvindn05: the openstack4j code already does this for the list calls for most part. This is just a additional check on our side for the log message.
For other getPort or getXXX() calls we need to handle null since that is the return value from openstack4j in case of any error or if the entity is not found.

emanoelxavier: I see so we dont need to do null checks on the caller side? If there is a bug on their library or somehow this changes I still think we should add emptyIfnull here though this way the control is on our code. For APIs returning a single object it makes sense to be null if not found.

arvindn05: We will end up with a lot of return types which looks like this....not sure if its worth it. I can update if you feel strongly that this would help going forward...let me know
(Set<String>) CollectionUtils.emptyIfNull(this.regions)
(List<? extends Server>) CollectionUtils.emptyIfNull(serverList)

arvindn05:
@karimull @larkinscarvalho Please let me know if you see anything which needs to be changed.

PR #487, opened 2017-09-22
Update OSC code for SDN controller API changes

arvindn05:
Updated OSC code for SDN Controller Changes https://github.com/opensecuritycontroller/sdn-controller-api/pull/36

Added inspectionHookId to vm port. This will store the flow classifier in case of networking sfc

Updated the tasks to remove checks for port group support if they are not applicable
Updated tasks to add placeholders for sfc calls once we have the entities defined on osc

**The build fails because the SDN controller api changes are not checked in.** We need to checkin these and api and plugins changes at the same time. once this code is approved, i will start the series of checkins

### DIFF
--- a/osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/securitygroup/VmPortHookCreateTask.java
+++ b/osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/securitygroup/VmPortHookCreateTask.java
@@ -16,7 +16,6 @@
  *******************************************************************************/
 package org.osc.core.broker.service.tasks.conformance.openstack.securitygroup;
 
-import java.util.Arrays;
 import java.util.Set;
 
 import javax.persistence.EntityManager;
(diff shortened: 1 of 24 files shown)


arvindn05:
@emanoelxavier please look over this change. we can discuss on a call.

osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/securitygroup/VmPortHookCreateTask.java
```diff
@@ -90,21 +88,10 @@ public void executeTransaction(EntityManager em) throws Exception {
...
-                            this.securityGroupInterface.getTagValue(), encapsulationType,
-                            this.securityGroupInterface.getOrder(), FailurePolicyType.valueOf(this.securityGroupInterface.getFailurePolicyType().name()));
-                }
-            } else {
-                controller.installInspectionHook(Arrays.asList(new NetworkElementImpl(this.vmPort)),
-                        new DefaultInspectionPort(ingressPort, egressPort, null),
-                        this.securityGroupInterface.getTagValue(), encapsulationType,
-                        this.securityGroupInterface.getOrder(), FailurePolicyType.valueOf(this.securityGroupInterface.getFailurePolicyType().name()));
-            }
+            controller.installInspectionHook(new NetworkElementImpl(this.vmPort),
```
emanoelxavier: Who is now running the previous code?

PR #490, opened 2017-09-24
K8s sync sg label tasks

dmitryintel:
So far test data only


emanoelxavier:
@dmitryintel I am picking this up now from this point. I will send a new PR including these commits and the rest. Thx.

emanoelxavier:
Included on #491

PR #491, opened 2017-09-25
Adding tasks and metatasks to syncronize k8s security group members

emanoelxavier:


emanoelxavier:
This includes #490

osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/k8s/securitygroup/CreateK8sLabelPodTask.java
```diff
@@ -0,0 +1,121 @@
...
+        Pod existingPod = PodEntityMgr.findExternalId(em, this.k8sPod.getUid());
+        SecurityGroup thisSG = this.label.getSecurityGroupMembers().iterator().next().getSecurityGroup();
+
+        if (existingPod != null) {
+            for (Label existingPodLabel : existingPod.getLabels()) {
+                if (existingPodLabel.getMarkedForDeletion() == true) {
+                    continue;
+                }
+
+                for (SecurityGroupMember sgm : existingPodLabel.getSecurityGroupMembers()) {
```
dmitryintel: Say the sgm is not a part of thisSG. Who could have marked it for deletion. Is it still locked by that other task right now? And if not, does that mean the other task has completed without actually deleting

emanoelxavier: It could be part of another SG that is either in the process of deletion of whose deletion failed to finish. The user could have triggered the deletion.

osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/k8s/securitygroup/DeleteK8sLabelPodTask.java
```diff
@@ -0,0 +1,77 @@
...
+        OSCEntityManager<Pod> dsEmgr = new OSCEntityManager<Pod>(Pod.class, em, this.txBroadcastUtil);
+        this.pod = dsEmgr.findByPrimaryKey(this.pod.getId());
+
+        // If the pod contains any label that is not the same as this label, do not delete the pod. NOOP
+        if (this.pod.getLabels().stream().anyMatch(label -> !label.getValue().equals(this.label.getValue()))) {
+            LOG.info(String.format("The pod %s is associated with another security group member. Skipping deletion.", this.pod.getName()));
+            return;
+        }
+
+        OSCEntityManager.delete(em, this.pod.getPorts().iterator().next(), this.txBroadcastUtil);
```
dmitryintel: This here assumes a single podPort per pod. Works in this setup, but an assumption it is

emanoelxavier: Correct, currently k8s only supports one port.

osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/securitygroup/SecurityGroupMemberDeleteTask.java
```diff
@@ -75,7 +78,24 @@ public void executeTransaction(EntityManager em) throws Exception {
             }
             this.log.info("Deleting Security Group member from " + this.sgm.getSecurityGroup().getName());
 
-        } else if (this.sgm.getType() == SecurityGroupMemberType.NETWORK) {
+        } else if (this.sgm.getType() == SecurityGroupMemberType.LABEL) {
+            Label label = this.sgm.getLabel();
+
+            if (label.getSecurityGroupMembers().size() == 1) {
+                this.log.info("No other references to Label found. Deleting Label " + label.getValue());
+                for (Pod pod : label.getPods()) {
```
dmitryintel: What if pod is associated to another unprotected label?

PR #499, opened 2017-09-28
Create packages for SFC code and refactor network element

arvindn05:
Create packages for SFC code
Create update service function chain task

Refactor code so folks can add code in right places


arvindn05:
@emanoelxavier please look at this since i did make changes to delete podnetwork impl and other port group related items

PR #500, opened 2017-09-28
Using the correct domain when deleting the port group

emanoelxavier:


emanoelxavier:
@dmitryintel please take a look at this PR when you have a chance.

PR #506, opened 2017-09-29
Update sfc task and unit test

arvindn05:


osc-server/src/test/java/org/osc/core/broker/service/tasks/conformance/openstack/sfc/UpdateServiceFunctionChainTaskTest.java
```diff
@@ -16,17 +16,170 @@
...
+        Mockito.when(this.em.find(ServiceFunctionChain.class, this.sfc.getId())).thenReturn(this.sfc);
+
+        Mockito.when(this.apiFactoryServiceMock.createNetworkRedirectionApi(mockVc)).thenReturn(this.sdnApi);
+
+    }
+
     @Test
-    public void test() {
-        fail("Not yet implemented");
+    public void testExecuteTransaction_WithSecurityGroupSFCIdSetToNull_ExpectUpdate() throws Exception {
```
emanoelxavier: testExecutedTransaction... -> testExecute...

osc-server/src/test/java/org/osc/core/broker/service/tasks/conformance/openstack/sfc/UpdateServiceFunctionChainTaskTest.java
```diff
@@ -16,17 +16,170 @@
...
+        // Arrange
+        UpdateServiceFunctionChainTask updateTask = this.task.create(this.sfc, this.sg, this.networkElementList);
+        Mockito.when(this.sdnApi
+                .updateNetworkElement(argThat(new NetworkElementIdMatcher(this.sg.getNetworkElementId())), any()))
+                .thenReturn(new NetworkElementImpl("UPDATED_SFC_ID"));
+
+        assertNull(this.sg.getNetworkElementId());
```
emanoelxavier: no need for this assertion, this is basically testing the test code which is redundant.

osc-server/src/test/java/org/osc/core/broker/service/tasks/conformance/openstack/sfc/UpdateServiceFunctionChainTaskTest.java
```diff
@@ -16,17 +16,170 @@
...
+
+        // Act.
+        updateTask.execute();
+
+        // Assert
+        assertEquals(this.sg.getNetworkElementId(), "UPDATED_SFC_ID");
+    }
+
+    @Test
+    public void testExecuteTransaction_WithSecurityGroupSFCIdSetToNotEmpty_ExpectUpdate() throws Exception {
```
emanoelxavier: testExecute...

osc-server/src/test/java/org/osc/core/broker/service/tasks/conformance/openstack/sfc/UpdateServiceFunctionChainTaskTest.java
```diff
@@ -16,17 +16,170 @@
...
+                .thenReturn(new NetworkElementImpl("UPDATED_SFC_ID"));
+        // Act.
+        updateTask.execute();
+
+        // Assert
+        assertEquals(this.sg.getNetworkElementId(), "UPDATED_SFC_ID");
+    }
+
+    @Test
+    public void testExecuteTransaction_CallsSDNUpdate_ExpectUpdate() throws Exception {
```
emanoelxavier: I thought all tests called SDNUpdate. What is the intent of this test? Maybe make the name match better.

osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/sfc/UpdateServiceFunctionChainTask.java
```diff
@@ -56,18 +58,25 @@ public UpdateServiceFunctionChainTask create(ServiceFunctionChain sfc, SecurityG
...
         this.sfc = em.find(ServiceFunctionChain.class, this.sfc.getId());
         this.securityGroup = em.find(SecurityGroup.class, this.securityGroup.getId());
-        this.updatedPortPairGroups.size();
-        //TODO: arvind implement this
 
+        try (SdnRedirectionApi sdnApi = this.apiFactory
+                .createNetworkRedirectionApi(this.securityGroup.getVirtualizationConnector())) {
+            NetworkElement updatedChain = sdnApi.updateNetworkElement(
+                    new NetworkElementImpl(this.securityGroup.getNetworkElementId()), this.updatedPortPairGroups);
+            if (!updatedChain.getElementId().equals(this.securityGroup.getNetworkElementId())) {
```
larkinscarvalho: Why do we check the updated chain ID? Do we expect a new sfc entity being created on Openstack as a part of updateNetworkElement method?

osc-server/src/test/java/org/osc/core/broker/service/tasks/conformance/openstack/sfc/UpdateServiceFunctionChainTaskTest.java
arvindn05: making sure that the task execution actually updates a null field...if something does change in the testInitialize we are making sure the update still happens correctly

osc-server/src/test/java/org/osc/core/broker/service/tasks/conformance/openstack/sfc/UpdateServiceFunctionChainTaskTest.java
arvindn05: changed to CallsSDNUpdateWithProvidedParameters

osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/sfc/UpdateServiceFunctionChainTask.java
arvindn05: yes. It really depends on the implementation though.  some plugins can consider SFC to be immutable and only way to update is to delete/recreate.

We could get an updated chain id or the existing chain id. Either case, we need to make sure our DB reflects the right values.

osc-server/src/test/java/org/osc/core/broker/service/tasks/conformance/openstack/sfc/UpdateServiceFunctionChainTaskTest.java
emanoelxavier: IMO the test code should be able to trust the test code, making assumptions in the test code should be ok and it keeps each method simple and targeted. As long as we keep dependencies light and most things local trusting it should be ok, cross checking becomes a necessity if the test code has many non local dependencies. If we cannot trust the test code itself does it mean we always need to double check the initialize? It could get complicated. Also, ideally all tests can follow the recipe arrange, act, assert. If we add asserts in the arrange part it can get confusing.

emanoelxavier: Another way to make the test more explicit while removing the assert in the arrange phase would be to simply do this.sg.setNe...(null). So no matter what the initialize does this test is setting it to null, it may also make the test more explicit.

PR #507, opened 2017-10-01
 Added update service function chain unit test APIs to check virtual …

karimull:
…system order

### DIFF
--- a/osc-server/src/test/java/org/osc/core/broker/service/servicefunctionchain/UpdateServiceFunctionChainServiceTest.java
+++ b/osc-server/src/test/java/org/osc/core/broker/service/servicefunctionchain/UpdateServiceFunctionChainServiceTest.java
@@ -58,7 +58,7 @@ public class UpdateServiceFunctionChainServiceTest extends BaseServiceFunctionCh
 	}
 
 	@Test
-	public void testDispatch_WithValidVirtualSystemList_SFCSIsUpdated() throws Exception {
+	public void testDispatch_WhenVirtualSystemListEmpty_SfcIsUpdated() throws Exception {
 
 		// Arrange
 		BaseDto dto = new BaseDto();
@@ -67,23 +67,23 @@ public class UpdateServiceFunctionChainServiceTest extends BaseServiceFunctionCh
 		request.setDto(dto);
 		request.setName(this.sfc.getName());
 		List<Long> vsIds = new ArrayList<Long>();
-
-		vsIds.add(this.vs.getId());
-		vsIds.add(this.vs1.getId());
-		request.setVirtualSystemIds(vsIds);
+		
 		Mockito.when(this.service.validator.validateAndLoad(request)).thenReturn(this.sfc);
 
 		// Act.
 		BaseJobResponse response = this.service.dispatch(request);
+		List<Long> sfcVsIdList = this.em.find(ServiceFunctionChain.class, this.sfc.getId()).getVirtualSystems().stream()
+                .map(vss -> vss.getId()).collect(Collectors.toList());
 
 		// Assert
+		Assert.assertEquals("The list of virtual system ids is not Empty", vsIds, sfcVsIdList);
 		Assert.assertNotNull("The returned response should not be null.", response);
 		Assert.assertEquals("The return Id should be equal to sfc Id", response.getId(), this.sfc.getId());
 	}
 
 
 	@Test
-	public void testDispatch_WithValidUpdate_SFCVirtualSystemListInOrder() throws Exception {
+	public void testDispatch_WhenVirtualSystemListNotEmpty_SfcIsUpdatedAndListInOrder() throws Exception {
 
 		// Arrange
 		BaseDto dto = new BaseDto();
--- a/osc-server/src/test/java/org/osc/core/broker/service/servicefunctionchain/UpdateServiceFunctionChainServiceTest.java
+++ b/osc-server/src/test/java/org/osc/core/broker/service/servicefunctionchain/UpdateServiceFunctionChainServiceTest.java
@@ -19,6 +19,7 @@ package org.osc.core.broker.service.servicefunctionchain;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import org.junit.Assert;
 import org.junit.Before;
@@ -27,6 +28,7 @@ import org.junit.runner.RunWith;
 import org.mockito.InjectMocks;
 import org.mockito.Mockito;
 import org.mockito.runners.MockitoJUnitRunner;
+import org.osc.core.broker.model.entities.virtualization.ServiceFunctionChain;
 import org.osc.core.broker.service.dto.BaseDto;
 import org.osc.core.broker.service.request.AddOrUpdateServiceFunctionChainRequest;
 import org.osc.core.broker.service.response.BaseJobResponse;
@@ -35,9 +37,9 @@ import org.osc.core.broker.service.response.BaseJobResponse;
 public class UpdateServiceFunctionChainServiceTest extends BaseServiceFunctionChainServiceTest {
 	@InjectMocks
 	private UpdateServiceFunctionChainService service;
-	
+
 	AddOrUpdateServiceFunctionChainRequest request;
-	
+
 	@Override
 	@Before
 	public void testInitialize() throws Exception {
@@ -56,40 +58,59 @@ public class UpdateServiceFunctionChainServiceTest extends BaseServiceFunctionCh
 	}
 
 	@Test
-	public void testDispatch_WithNonEmptyListOfVsIdsUpdateValidationSuccessful_ValidationSucceeds() throws Exception {
-		
+	public void testDispatch_WithValidVirtualSystemList_SFCSIsUpdated() throws Exception {
+
+		// Arrange
 		BaseDto dto = new BaseDto();
 		dto.setId(this.sfc.getId());
 		dto.setParentId(this.vc.getId());
 		request.setDto(dto);
 		request.setName(this.sfc.getName());
 		List<Long> vsIds = new ArrayList<Long>();
+
 		vsIds.add(this.vs.getId());
 		vsIds.add(this.vs1.getId());
 		request.setVirtualSystemIds(vsIds);
 		Mockito.when(this.service.validator.validateAndLoad(request)).thenReturn(this.sfc);
+
 		// Act.
 		BaseJobResponse response = this.service.dispatch(request);
+
+		// Assert
 		Assert.assertNotNull("The returned response should not be null.", response);
+		Assert.assertEquals("The return Id should be equal to sfc Id", response.getId(), this.sfc.getId());
 	}
-	
+
+
 	@Test
-	public void testDispatch_WhenVsIdsListUpdateValidationSuccessful_ValidationSucceeds() throws Exception {
-		
+	public void testDispatch_WithValidUpdate_SFCVirtualSystemListInOrder() throws Exception {
+
+		// Arrange
 		BaseDto dto = new BaseDto();
 		dto.setId(this.sfc.getId());
 		dto.setParentId(this.vc.getId());
 		request.setDto(dto);
 		request.setName(this.sfc.getName());
 		List<Long> vsIds = new ArrayList<Long>();
+
 		vsIds.add(this.vs1.getId());
 		vsIds.add(this.vs.getId());
 		request.setVirtualSystemIds(vsIds);
 		Mockito.when(this.service.validator.validateAndLoad(request)).thenReturn(this.sfc);
+
 		// Act.
+		// update sfc with virtual system ids list {2,1}
 		BaseJobResponse response = this.service.dispatch(request);
+
+		// check Virtual system is as updated
+		List<Long> sfcVsIdList = this.em.find(ServiceFunctionChain.class, this.sfc.getId()).getVirtualSystems().stream()
+				.map(vss -> vss.getId()).collect(Collectors.toList());
+		// read back sfc and cross check with local Virtual system list for
+		// order
+
+		// Assert
+		Assert.assertEquals("The list of virtual system ids is different than expected", vsIds, sfcVsIdList);
 		Assert.assertNotNull("The returned response should not be null.", response);
+		Assert.assertEquals("The return Id should be equal to sfc Id", response.getId(), this.sfc.getId());
 	}
-	
-	//TODO : karimull  Add more test cases in looking for updated in order. Need some ramp up here
 }


osc-server/src/test/java/org/osc/core/broker/service/servicefunctionchain/UpdateServiceFunctionChainServiceTest.java
```diff
@@ -82,14 +84,67 @@ public void testDispatch_WhenVsIdsListUpdateValidationSuccessful_ValidationSucce
...
+		request.setVirtualSystemIds(vsIds);
+		Mockito.when(this.service.validator.validateAndLoad(request)).thenReturn(this.sfc);
+		// Act.
+		BaseJobResponse response = this.service.dispatch(request);
+		Assert.assertNotNull("The returned response should not be null.", response);
+	}
+	
+	@Test
+	public void testDispatch_sfcVirutalSystemIdsInOrder_ValidationSucceeds() throws Exception {
+
```
emanoelxavier: // Arrange.

osc-server/src/test/java/org/osc/core/broker/service/servicefunctionchain/UpdateServiceFunctionChainServiceTest.java
```diff
@@ -82,14 +84,67 @@ public void testDispatch_WhenVsIdsListUpdateValidationSuccessful_ValidationSucce
...
+		dto.setId(this.sfc.getId());
+		dto.setParentId(this.vc.getId());
+		request.setDto(dto);
+		request.setName(this.sfc.getName());
+		List<Long> vsIds = new ArrayList<Long>();
 		vsIds.add(this.vs1.getId());
 		vsIds.add(this.vs.getId());
 		request.setVirtualSystemIds(vsIds);
 		Mockito.when(this.service.validator.validateAndLoad(request)).thenReturn(this.sfc);
 		// Act.
```
emanoelxavier: add a break line

osc-server/src/test/java/org/osc/core/broker/service/servicefunctionchain/UpdateServiceFunctionChainServiceTest.java
```diff
@@ -82,14 +84,67 @@ public void testDispatch_WhenVsIdsListUpdateValidationSuccessful_ValidationSucce
...
+		//update sfc with virtual system ids list {2,1}
```
emanoelxavier: // Assert. The AAA comments really help to review and keep the longer tests like this organized.

osc-server/src/test/java/org/osc/core/broker/service/servicefunctionchain/UpdateServiceFunctionChainServiceTest.java
```diff
@@ -82,14 +84,67 @@ public void testDispatch_WhenVsIdsListUpdateValidationSuccessful_ValidationSucce
...
 		BaseJobResponse response = this.service.dispatch(request);
+		
+		//check Virtual system is as updated
+		List<Long> sfcVsIdList = this.em.find(ServiceFunctionChain.class, this.sfc.getId()).
+				getVirtualSystems().stream().map(vss -> vss.getId()).collect(Collectors.toList());
+		//read back sfc and cross check with local Virtual system list for order
+		Assert.assertEquals(vsIds, sfcVsIdList);
```
emanoelxavier: Please ensure to use a string message in all your asserts assertEquals("The list of virtual system ids is different than expected", ...). When the test fails this helps to quickly diagnose.  Also look at other asserts.

emanoelxavier: Does this check the elements in the list?

osc-server/src/test/java/org/osc/core/broker/service/servicefunctionchain/UpdateServiceFunctionChainServiceTest.java
```diff
@@ -82,14 +84,67 @@ public void testDispatch_WhenVsIdsListUpdateValidationSuccessful_ValidationSucce
...
+		vsIds.add(this.vs1.getId());
+		request.setVirtualSystemIds(vsIds);
+		Mockito.when(this.service.validator.validateAndLoad(request)).thenReturn(this.sfc);
+		// Act.
+		BaseJobResponse response = this.service.dispatch(request);
+		Assert.assertNotNull("The returned response should not be null.", response);
+	}
+	
+	@Test
+	public void testDispatch_sfcVirutalSystemIdsInOrder_ValidationSucceeds() throws Exception {
```
emanoelxavier: I believe this is testing the service not the validation...then maybe ...validSFCVirtualSystemIds_SFCIsUpdated

osc-server/src/test/java/org/osc/core/broker/service/servicefunctionchain/UpdateServiceFunctionChainServiceTest.java
```diff
@@ -82,14 +84,67 @@ public void testDispatch_WhenVsIdsListUpdateValidationSuccessful_ValidationSucce
...
+		Mockito.when(this.service.validator.validateAndLoad(request)).thenReturn(this.sfc);
+		// Act.
+		//update sfc with virtual system ids list {2,1}
+		BaseJobResponse response = this.service.dispatch(request);
+		
+		//check Virtual system is as updated
+		List<Long> sfcVsIdList = this.em.find(ServiceFunctionChain.class, this.sfc.getId()).
+				getVirtualSystems().stream().map(vss -> vss.getId()).collect(Collectors.toList());
+		
+		//change the order to compare
```
emanoelxavier: why is this needed?

osc-server/src/test/java/org/osc/core/broker/service/servicefunctionchain/UpdateServiceFunctionChainServiceTest.java
```diff
@@ -82,14 +84,67 @@ public void testDispatch_WhenVsIdsListUpdateValidationSuccessful_ValidationSucce
...
 		// Act.
+		//update sfc with virtual system ids list {2,1}
 		BaseJobResponse response = this.service.dispatch(request);
+		
+		//check Virtual system is as updated
+		List<Long> sfcVsIdList = this.em.find(ServiceFunctionChain.class, this.sfc.getId()).
+				getVirtualSystems().stream().map(vss -> vss.getId()).collect(Collectors.toList());
+		//read back sfc and cross check with local Virtual system list for order
+		Assert.assertEquals(vsIds, sfcVsIdList);
 		Assert.assertNotNull("The returned response should not be null.", response);
```
emanoelxavier: Can you also check something about the content of the response? Maybe the expected id?

osc-server/src/test/java/org/osc/core/broker/service/servicefunctionchain/UpdateServiceFunctionChainServiceTest.java
```diff
@@ -82,14 +84,67 @@ public void testDispatch_WhenVsIdsListUpdateValidationSuccessful_ValidationSucce
...
 	}
 	
-	//TODO : karimull  Add more test cases in looking for updated in order. Need some ramp up here
+	@Test
+	public void testDispatch_sfcVirutalSystemIdsNotInOrder_ValidationFails() throws Exception {
```
emanoelxavier: This name does not align with the test code. The validator is not mocked to fail and the sfc seems to be updated. So in essence this tests looks like the previous one.

osc-server/src/test/java/org/osc/core/broker/service/servicefunctionchain/UpdateServiceFunctionChainServiceTest.java
karimull: Yes it is checking all the list elements.

osc-server/src/test/java/org/osc/core/broker/service/servicefunctionchain/UpdateServiceFunctionChainServiceTest.java
karimull: Agreed , it is kind of similar to previous one, but thought of checking in reverse also. 
May not needed ..taking it off.

osc-server/src/test/java/org/osc/core/broker/service/servicefunctionchain/UpdateServiceFunctionChainServiceTest.java
karimull: removed the test

osc-server/src/test/java/org/osc/core/broker/service/servicefunctionchain/UpdateServiceFunctionChainServiceTest.java
karimull: added check

osc-server/src/test/java/org/osc/core/broker/service/servicefunctionchain/UpdateServiceFunctionChainServiceTest.java
```diff
@@ -60,91 +60,77 @@ public void testDispatch_WithNullRequest_ThrowsNullPointerException() throws Exc
 	@Test
 	public void testDispatch_WithNonEmptyListOfVsIdsUpdateValidationSuccessful_ValidationSucceeds() throws Exception {
 		
+		//Arrange
```
emanoelxavier: picky `//Arrange` -> `// Arrange.` Observe the space and period. Also check other places.

osc-server/src/test/java/org/osc/core/broker/service/servicefunctionchain/UpdateServiceFunctionChainServiceTest.java
```diff
@@ -60,91 +60,77 @@ public void testDispatch_WithNullRequest_ThrowsNullPointerException() throws Exc
 	@Test
 	public void testDispatch_WithNonEmptyListOfVsIdsUpdateValidationSuccessful_ValidationSucceeds() throws Exception {
```
emanoelxavier: The name is not accurate. You are not testing that the validation succeeds, the validator is mocked it should have its own tests. You are testing that **when** the validation succeeds the SFC is updated. This names gives the impression you are testing the validation which is being mocked. Maybe something like testDispatch_WithNonEmpty..._SFCIsUpdated

osc-server/src/test/java/org/osc/core/broker/service/servicefunctionchain/UpdateServiceFunctionChainServiceTest.java
```diff
@@ -60,91 +60,77 @@ public void testDispatch_WithNullRequest_ThrowsNullPointerException() throws Exc
...
 		vsIds.add(this.vs.getId());
 		vsIds.add(this.vs1.getId());
 		request.setVirtualSystemIds(vsIds);
 		Mockito.when(this.service.validator.validateAndLoad(request)).thenReturn(this.sfc);
+		
 		// Act.
 		BaseJobResponse response = this.service.dispatch(request);
+		
+		//Assert
 		Assert.assertNotNull("The returned response should not be null.", response);
```
emanoelxavier: Also please validate what should be in the response.

osc-server/src/test/java/org/osc/core/broker/service/servicefunctionchain/UpdateServiceFunctionChainServiceTest.java
```diff
@@ -60,91 +60,77 @@ public void testDispatch_WithNullRequest_ThrowsNullPointerException() throws Exc
...
-		//update sfc with virtual system ids list {2,1}
 		BaseJobResponse response = this.service.dispatch(request);
 		
-		//check Virtual system is as updated
-		List<Long> sfcVsIdList = this.em.find(ServiceFunctionChain.class, this.sfc.getId()).
-				getVirtualSystems().stream().map(vss -> vss.getId()).collect(Collectors.toList());
-		//read back sfc and cross check with local Virtual system list for order
-		Assert.assertEquals(vsIds, sfcVsIdList);
+		//Assert
 		Assert.assertNotNull("The returned response should not be null.", response);
```
emanoelxavier: Again, assert what should be in the response. Also how do we know that the service is calling the db with the correct info? If there is a regression on that code this is not going to catch it.

osc-server/src/test/java/org/osc/core/broker/service/servicefunctionchain/UpdateServiceFunctionChainServiceTest.java
```diff
@@ -60,91 +60,77 @@ public void testDispatch_WithNullRequest_ThrowsNullPointerException() throws Exc
...
-		
-		//change the order to compare
-		vsIds.clear();
-		vsIds.add(this.vs.getId());
-		vsIds.add(this.vs1.getId());
 		//read back sfc and cross check with local Virtual system list for order
-		Assert.assertNotEquals(vsIds, sfcVsIdList);
+		
+		//Assert
+		Assert.assertEquals("The list of virtual system ids is different than expected", vsIds, sfcVsIdList);
```
emanoelxavier: Is this test the same as the previous one but with additional validation? If so you should have just a single test case containing all the validation/asserts that are applicable. You would have a new test case only if the inputs, action or mocked state changes but you can pack as much validation as possible on the same test method/case

karimull: Added three test cases with Empty list of Virtual systems, a updated virtual systems and last one is Validation of the order. Yes the last two are similar, but do not want to stub into everything as I change the input list to check back the action.

osc-server/src/test/java/org/osc/core/broker/service/servicefunctionchain/UpdateServiceFunctionChainServiceTest.java
karimull: added response id check

osc-server/src/test/java/org/osc/core/broker/service/servicefunctionchain/UpdateServiceFunctionChainServiceTest.java
karimull: added response id check

karimull:
Removed one redundant test case and updated names

osc-server/src/test/java/org/osc/core/broker/service/servicefunctionchain/UpdateServiceFunctionChainServiceTest.java
```diff
@@ -56,40 +58,59 @@ public void testDispatch_WithNullRequest_ThrowsNullPointerException() throws Exc
...
-	
+
+
 	@Test
-	public void testDispatch_WhenVsIdsListUpdateValidationSuccessful_ValidationSucceeds() throws Exception {
-		
+	public void testDispatch_WhenVirtualSystemListNotEmpty_SfcIsUpdatedAndListInOrder() throws Exception {
+
+		// Arrange
 		BaseDto dto = new BaseDto();
```
arvindn05: since you are doing the same in multiple places can you just do this in testInitialize?

PR #508, opened 2017-10-03
Fixing sync of k8s label security member

emanoelxavier:


emanoelxavier:
1. Improves the names of some the new tasks and metatasks.
2. Conforms the port group after previous tasks are completed not necessarily succeeded.
3. Appends instead of adding the pod create tasks due to a transactional error when multiple pods on the same label are sync'd at the same time.

osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/k8s/securitygroup/CheckK8sSecurityGroupLabelMetaTask.java
```diff
@@ -53,8 +53,8 @@ public TaskGraph getTaskGraph() {
 
     @Override
     public String getName() {
-        return String.format("Conformance: %s Kubernetes Security Group Member Label %s",
-                this.isDelete ? "Delete" : "Check", this.sgm.getLabel().getName());
+        return String.format("%s Kubernetes Security Group Member Label %s",
```
arvindn05: you can remove kubernetes here.

Checking Security group member label ...

osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/k8s/securitygroup/CreateK8sLabelPodTask.java
```diff
@@ -125,7 +125,7 @@ public CreateK8sLabelPodTask create(KubernetesPod k8sPod, Label label) {
 
     @Override
     public String getName() {
-        return String.format("Creating the pod id %s, name %s", this.k8sPod.getUid(), this.k8sPod.getName());
+        return String.format("Persisting the pod name %s, id %s under the member %s", this.k8sPod.getName(), this.k8sPod.getUid(), this.label.getValue());
```
arvindn05: nit pick. Generally for id's or name we put them in between ' '

For example
>Deleting Server 'cirros-1nic-1-3' from region 'RegionOne'

osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/k8s/securitygroup/UpdateK8sSecurityGroupMemberLabelMetaTask.java
```diff
@@ -89,7 +89,9 @@ public void executeTransaction(EntityManager em) throws Exception {
         Set<Pod> dbPodsToDelete = emptyIfNull(label.getPods()).stream()
                 .filter(p -> !existingPodIdsInK8s.contains(p.getExternalId())).collect(Collectors.toSet());
 
-        k8sPodsToCreate.forEach(p -> this.tg.addTask(this.labelPodCreateTask.create(p, this.sgm.getLabel())));
+        // TODO emanoel: Using append instead of adding here because of a transactional issue updating multiple
+        // pods in the same label concurrently. An alternative to this might be locking the label member entity.
+        k8sPodsToCreate.forEach(p -> this.tg.appendTask(this.labelPodCreateTask.create(p, this.sgm.getLabel())));
         dbPodsToDelete.forEach(p -> this.tg.addTask(this.labelPodDeleteTask.create(p, this.sgm.getLabel())));
```
arvindn05: would this impact delete the same way? should this be append as well?

osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/k8s/securitygroup/UpdateOrDeleteK8sSecurityGroupMetaTask.java
```diff
@@ -66,7 +67,7 @@ public TaskGraph getTaskGraph() {
     @Override
     public String getName() {
         final boolean isDelete = this.sg.getMarkedForDeletion();
-        return String.format("%s Kubernetes Security Group %s", isDelete ? "Delete" : "Update",
+        return String.format("%s Kubernetes Security Group %s", isDelete ? "Deleting" : "Updating",
```
arvindn05: remove kubernetes

osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/k8s/securitygroup/UpdateK8sSecurityGroupMemberLabelMetaTask.java
emanoelxavier: The deletion is not updating the label entity, just deleting the pod and port so i dont think it will have the same impact, I will cover that by deleting two pods on the same label and syncing to make sure. If there is an issue I will update the PR with the same workaround before merging.

osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/k8s/securitygroup/CheckK8sSecurityGroupLabelMetaTask.java
```diff
@@ -53,8 +53,8 @@ public TaskGraph getTaskGraph() {
 
     @Override
     public String getName() {
-        return String.format("Conformance: %s Kubernetes Security Group Member Label %s",
-                this.isDelete ? "Delete" : "Check", this.sgm.getLabel().getName());
+        return String.format("%s Kubernetes Security Group Member Label '%s'",
```
arvindn05: @emanoelxavier still see "Kubernetes" here

PR #510, opened 2017-10-03
Check for Deleting DA which is binded to SFC, if  binded should throw an…

karimull:
… exception

### DIFF
--- a/osc-server/src/main/java/org/osc/core/broker/service/DeleteDistributedApplianceService.java
+++ b/osc-server/src/main/java/org/osc/core/broker/service/DeleteDistributedApplianceService.java
@@ -120,7 +120,7 @@ implements DeleteDistributedApplianceServiceApi {
     @Override
     public BaseJobResponse exec(BaseDeleteRequest request, EntityManager em) throws Exception {
         if (this.validator == null) {
-            this.validator = new DeleteDistributedApplianceRequestValidator(em);
+            this.validator = new DeleteDistributedApplianceRequestValidator(em, apiFactoryService);
         }
 
         UnlockObjectMetaTask ult = null;
--- a/osc-server/src/main/java/org/osc/core/broker/service/DeleteDistributedApplianceService.java
+++ b/osc-server/src/main/java/org/osc/core/broker/service/DeleteDistributedApplianceService.java
@@ -120,7 +120,7 @@ implements DeleteDistributedApplianceServiceApi {
     @Override
     public BaseJobResponse exec(BaseDeleteRequest request, EntityManager em) throws Exception {
         if (this.validator == null) {
-            this.validator = new DeleteDistributedApplianceRequestValidator(em, apiFactoryService);
+            this.validator = new DeleteDistributedApplianceRequestValidator(em);
         }
 
         UnlockObjectMetaTask ult = null;
--- a/osc-server/src/main/java/org/osc/core/broker/service/validator/DeleteDistributedApplianceRequestValidator.java
+++ b/osc-server/src/main/java/org/osc/core/broker/service/validator/DeleteDistributedApplianceRequestValidator.java
@@ -59,13 +59,10 @@ public class DeleteDistributedApplianceRequestValidator implements RequestValida
         }
         
         for (VirtualSystem vs : da.getVirtualSystems()) {
-            if (!this.apiFactoryService.supportsNeutronSFC(vs)) {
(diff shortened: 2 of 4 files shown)


osc-server/src/main/java/org/osc/core/broker/service/validator/DeleteDistributedApplianceRequestValidator.java
```diff
@@ -50,6 +57,18 @@ public DistributedAppliance validateAndLoad(BaseDeleteRequest request) throws Ex
                             + request.getId()
                             + " is not marked for deletion and force delete operation is applicable only for entries marked for deletion.");
         }
+        
+        for (VirtualSystem vs : da.getVirtualSystems()) {
+            if (!this.apiFactoryService.supportsNeutronSFC(vs)) {
```
arvindn05: we dont care about neutron sfc support. If the vs has a sfc attached to it, we can throw the exception

osc-server/src/main/java/org/osc/core/broker/service/validator/DeleteDistributedApplianceRequestValidator.java
```diff
@@ -50,6 +57,18 @@ public DistributedAppliance validateAndLoad(BaseDeleteRequest request) throws Ex
...
+             // if first virtual system is not neutronSFC break...
+                break; 
+            }
+            if (vs.getServiceFunctionChains().size() > 0) {
+                throw new VmidcBrokerValidationException("Distributed Appilance with ID " + request.getId()
```
arvindn05: can we make the exception message more explicit? specify the sfc name/id which they need to unbind and from which virtual system etc

Something along the lines of(without the abbreviations)
Cannot delete DA as one of the VS'es VSA is being referenced by SFCA

osc-server/src/main/java/org/osc/core/broker/service/validator/DeleteDistributedApplianceRequestValidator.java
```diff
@@ -59,13 +59,10 @@ public DistributedAppliance validateAndLoad(BaseDeleteRequest request) throws Ex
...
         for (VirtualSystem vs : da.getVirtualSystems()) {
-            if (!this.apiFactoryService.supportsNeutronSFC(vs)) {
-             // if first virtual system is not neutronSFC break...
-                break; 
-            }
             if (vs.getServiceFunctionChains().size() > 0) {
-                throw new VmidcBrokerValidationException("Distributed Appilance with ID " + request.getId()
-                        + " is binded to a service function chain, unbind to delete");
+                throw new VmidcBrokerValidationException("Cannot delete Distributed Appilance with ID " + request.getId()
+                        + " as its associated Virtual System : " + vs.getName() + " is beign referenced by Service Function Chain : " +
```
arvindn05: typo being

osc-server/src/main/java/org/osc/core/broker/service/DeleteDistributedApplianceService.java
```diff
@@ -120,7 +120,7 @@ public void completed(Job job) {
     @Override
     public BaseJobResponse exec(BaseDeleteRequest request, EntityManager em) throws Exception {
         if (this.validator == null) {
-            this.validator = new DeleteDistributedApplianceRequestValidator(em);
+            this.validator = new DeleteDistributedApplianceRequestValidator(em, apiFactoryService);
```
arvindn05: api factory service no longer used

osc-server/src/main/java/org/osc/core/broker/service/validator/DeleteDistributedApplianceRequestValidator.java
```diff
@@ -50,6 +57,15 @@ public DistributedAppliance validateAndLoad(BaseDeleteRequest request) throws Ex
                             + request.getId()
                             + " is not marked for deletion and force delete operation is applicable only for entries marked for deletion.");
         }
+        
+        for (VirtualSystem vs : da.getVirtualSystems()) {
+            if (vs.getServiceFunctionChains().size() > 0) {
+                throw new VmidcBrokerValidationException("Cannot delete Distributed Appilance with ID " + request.getId()
+                        + " as its associated Virtual System : " + vs.getName() + " is beign referenced by Service Function Chain : " +
```
arvindn05: typo in being

PR #515, opened 2017-10-06
Unit test cases for bindSecurityGroupServices

karimull:
More unit test cases and bug fixes in binding services found during unit test runs

Following bugs are fixed that were found during unit test

1.During a check for , size of VirtualSystem List given compare to size of SFC VirtualSystem list we throw an exception and in that message the parameters were given wrong.

2. Retrieving sfc entities per virtual system is done through DB query now. Earlier we were getting them through virtual system get servicefunction chain attribute, where as SFC were never set in virtual system list.

3.When checking for redundant SFC chains in the system, we are querying for virtual system of a given SFC is part of any other SFC. And If they exist another SFC with same VS and if a security group exist with this other SFC and same tenant , we call it a redundant chain. In this check we should not include the current SG that is configured as part of getting all SGs. Fix it.

4. Addressed one of the comment which was a bug, like conditional check in the code during sfc redundancy check

### DIFF
--- a/osc-server/src/main/java/org/osc/core/broker/service/persistence/ServiceFunctionChainEntityMgr.java
+++ b/osc-server/src/main/java/org/osc/core/broker/service/persistence/ServiceFunctionChainEntityMgr.java
@@ -20,6 +20,9 @@ import java.util.ArrayList;
 import java.util.List;
 
 import javax.persistence.EntityManager;
+import javax.persistence.criteria.CriteriaBuilder;
+import javax.persistence.criteria.CriteriaQuery;
+import javax.persistence.criteria.Root;
 
 import org.osc.core.broker.model.entities.appliance.VirtualSystem;
 import org.osc.core.broker.model.entities.virtualization.ServiceFunctionChain;
@@ -44,6 +47,17 @@ public class ServiceFunctionChainEntityMgr {
 
 	public static ServiceFunctionChain findById(EntityManager em, Long id) {
         return em.find(ServiceFunctionChain.class, id);
+    }
+	
+    public static List<ServiceFunctionChain> listServiceFunctionChainsByVirtualSystem(EntityManager em, VirtualSystem vs) {
+        CriteriaBuilder cb = em.getCriteriaBuilder();
+
+        CriteriaQuery<ServiceFunctionChain> query = cb.createQuery(ServiceFunctionChain.class);
+
+        Root<ServiceFunctionChain> root = query.from(ServiceFunctionChain.class);
+        query = query.select(root).where(cb.equal(root.join("virtualSystems"), vs)).orderBy(cb.asc(root.get("name")));
+
+        return em.createQuery(query).getResultList();
     }
 
 }
--- a/osc-server/src/main/java/org/osc/core/broker/service/persistence/ServiceFunctionChainEntityMgr.java
+++ b/osc-server/src/main/java/org/osc/core/broker/service/persistence/ServiceFunctionChainEntityMgr.java
@@ -48,14 +48,14 @@ public class ServiceFunctionChainEntityMgr {
 	public static ServiceFunctionChain findById(EntityManager em, Long id) {
         return em.find(ServiceFunctionChain.class, id);
     }
-	
+
     public static List<ServiceFunctionChain> listServiceFunctionChainsByVirtualSystem(EntityManager em, VirtualSystem vs) {
         CriteriaBuilder cb = em.getCriteriaBuilder();
 
         CriteriaQuery<ServiceFunctionChain> query = cb.createQuery(ServiceFunctionChain.class);
 
         Root<ServiceFunctionChain> root = query.from(ServiceFunctionChain.class);
-        query = query.select(root).where(cb.equal(root.join("virtualSystems"), vs)).orderBy(cb.asc(root.get("name")));
+        query = query.select(root).where(cb.equal(root.join("virtualSystems"), vs));
 
         return em.createQuery(query).getResultList();
     }
--- a/osc-server/src/main/java/org/osc/core/broker/service/securitygroup/BindSecurityGroupService.java
+++ b/osc-server/src/main/java/org/osc/core/broker/service/securitygroup/BindSecurityGroupService.java
@@ -408,10 +408,10 @@ public class BindSecurityGroupService extends ServiceDispatcher<BindSecurityGrou
              
              //if this virtual system is pointing to more than one SFC and one of the SFC other than given SFC
              //is already binded to SG and active, throw exception;
-             //Get All the sfcs having this virutalsystem in the chain
+             //Get All the sfcs having this virtual system in the chain
              List<ServiceFunctionChain> serviceFunctionChainList = ServiceFunctionChainEntityMgr.
                                                                  listServiceFunctionChainsByVirtualSystem(em, vs);
-            if (serviceFunctionChainList != null && !serviceFunctionChainList.isEmpty()) {
+            if (serviceFunctionChainList != null) {
                 for (ServiceFunctionChain serviceFunctionChain : serviceFunctionChainList) {
                     if (!serviceFunctionChain.getId().equals(sfc.getId())) {
                         checkVirtualSystemRedundancyInServiceFunctionChains(em, serviceFunctionChain.getId(), vs);
--- a/osc-server/src/main/java/org/osc/core/broker/service/securitygroup/BindSecurityGroupService.java
+++ b/osc-server/src/main/java/org/osc/core/broker/service/securitygroup/BindSecurityGroupService.java
@@ -45,6 +45,7 @@ import org.osc.core.broker.service.persistence.OSCEntityManager;
 import org.osc.core.broker.service.persistence.PolicyEntityMgr;
 import org.osc.core.broker.service.persistence.SecurityGroupEntityMgr;
 import org.osc.core.broker.service.persistence.SecurityGroupInterfaceEntityMgr;
+import org.osc.core.broker.service.persistence.ServiceFunctionChainEntityMgr;
 import org.osc.core.broker.service.persistence.VirtualSystemEntityMgr;
 import org.osc.core.broker.service.persistence.VirtualizationConnectorEntityMgr;
 import org.osc.core.broker.service.request.BindSecurityGroupRequest;
@@ -70,7 +71,7 @@ public class BindSecurityGroupService extends ServiceDispatcher<BindSecurityGrou
 	private ConformService conformService;
 
 	@Reference
-	private ApiFactoryService apiFactoryService;
+	public ApiFactoryService apiFactoryService;
 
 	private SecurityGroup securityGroup;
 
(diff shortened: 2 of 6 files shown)


arvindn05:
@karimull please add description on what bug fixes

osc-server/src/main/java/org/osc/core/broker/service/securitygroup/BindSecurityGroupService.java
```diff
@@ -405,9 +406,12 @@ private ServiceFunctionChain validateAndLoadSfcBind(EntityManager em, BindSecuri
              VirtualSystem vs = validateAndLoadVirtualSystem(em, virtualSystemId);
              
              //if this virtual system is pointing to more than one SFC and one of the SFC other than given SFC
-             //is already binded to SG and active, throw exception
-            if (vs.getServiceFunctionChains().size() > 1) {
-                for (ServiceFunctionChain serviceFunctionChain : vs.getServiceFunctionChains()) {
+             //is already binded to SG and active, throw exception;
+             //Get All the sfcs having this virutalsystem in the chain
```
emanoelxavier: virtualsystem -> virtual system

osc-server/src/main/java/org/osc/core/broker/service/securitygroup/BindSecurityGroupService.java
```diff
@@ -405,9 +406,12 @@ private ServiceFunctionChain validateAndLoadSfcBind(EntityManager em, BindSecuri
...
+             List<ServiceFunctionChain> serviceFunctionChainList = ServiceFunctionChainEntityMgr.
+                                                                 listServiceFunctionChainsByVirtualSystem(em, vs);
+            if (serviceFunctionChainList != null && !serviceFunctionChainList.isEmpty()) {
```
emanoelxavier: !isEmpty is not needed. The loop wont happen if empty

osc-server/src/test/java/org/osc/core/broker/service/BindSecurityGroupServicePersistenDbTest.java
```diff
@@ -0,0 +1,521 @@
...
+import org.osc.core.broker.util.db.DBConnectionManager;
+import org.osc.core.common.virtualization.VirtualizationType;
+import org.osc.core.test.util.TestTransactionControl;
+import org.powermock.api.mockito.PowerMockito;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest(LockUtil.class)
+public class BindSecurityGroupServicePersistenDbTest {
```
emanoelxavier: BindSecurityGroupServiceWithPersistenceTest ?

osc-server/src/test/java/org/osc/core/broker/service/BindSecurityGroupServiceTest.java
```diff
@@ -0,0 +1,574 @@
...
+
+        // Act
+        this.bindSecurityGroupService.dispatch(request);
+
+    }
+
+    @Test
+    public void testValidate_WhenSgIsMarkedForDeletion_ThrowsInvalidRequestException() throws Exception {
+        // Arrange
+        BindSecurityGroupRequest request = createRequest(VALID_ID_MARK_FOR_DELETION, VALID_ID_MARK_FOR_DELETION,
```
emanoelxavier: From the tests it seems that the actual service, not  a  validator, is doing the validation. I think it should be a validator and the service tests would simply mock it. Or is this by design?

karimull:
Following bugs are fixed that were found during unit test

1. During a check for ,  size of VirtualSystem List given compare to size of SFC  VirtualSystem  list we throw an exception and in that message the parameters were give wrong.

2. When checking any existing SFC other than the given SFC to bind connected to a SG and particular Tenant(Redundancy chain check), was getting SFC list from Virtual Systems. But in Add or update of SFC entity services we were not updating SFC list in VirtualSystem Entity.  So added a query instead of getting from virtualsystem sfc entity list, which is more preferred.

3. When checking for redundant SFC chains in the system, we are querying for virtual system of a given SFC is part of any other SFC. And If they exist another SFC with same VS and if a security group exist with this other SFC and same tenant , we call it a redundant chain. In this check we should not include the current SG that is configured as part of getting all SGs. Fix it.

osc-server/src/main/java/org/osc/core/broker/service/securitygroup/BindSecurityGroupService.java
karimull: I wanted to check of list exist and not empty.  will query return just null if no data ?.
please let me know.

osc-server/src/test/java/org/osc/core/broker/service/BindSecurityGroupServicePersistenDbTest.java
karimull: will modify

osc-server/src/test/java/org/osc/core/broker/service/BindSecurityGroupServiceTest.java
karimull: Yes there no validator for the this service. Everything is done as part  of service. But in this case everything is mocked..no data base.

osc-server/src/main/java/org/osc/core/broker/service/securitygroup/BindSecurityGroupService.java
karimull: Will make this change if there is change request for this file..later.

osc-server/src/test/java/org/osc/core/broker/service/BindSecurityGroupServiceTest.java
emanoelxavier: I see there should be a validator but since this is a legacy service it is out of scope for your changes.

osc-server/src/main/java/org/osc/core/broker/service/securitygroup/BindSecurityGroupService.java
emanoelxavier: You can leave the null check but the empty check is not necessary. If empty the loop below will not run at all, it is already safe.

osc-server/src/main/java/org/osc/core/broker/service/securitygroup/BindSecurityGroupService.java
karimull: corrected the typo

osc-server/src/main/java/org/osc/core/broker/service/securitygroup/BindSecurityGroupService.java
karimull: removed empty check

osc-server/src/main/java/org/osc/core/broker/service/persistence/ServiceFunctionChainEntityMgr.java
```diff
@@ -45,5 +48,16 @@ public static void fromEntity(ServiceFunctionChain sfcEntity, ServiceFunctionCha
 	public static ServiceFunctionChain findById(EntityManager em, Long id) {
         return em.find(ServiceFunctionChain.class, id);
     }
+	
+    public static List<ServiceFunctionChain> listServiceFunctionChainsByVirtualSystem(EntityManager em, VirtualSystem vs) {
```
arvindn05: why do we need this? cant we just do vs.getServiceFunctionChains()?

is there a significance to sorting by name?

osc-server/src/main/java/org/osc/core/broker/service/securitygroup/BindSecurityGroupService.java
```diff
@@ -70,7 +71,7 @@
 	private ConformService conformService;
 
 	@Reference
-	private ApiFactoryService apiFactoryService;
+	public ApiFactoryService apiFactoryService;
```
arvindn05: can be package private

osc-server/src/main/java/org/osc/core/broker/service/securitygroup/BindSecurityGroupService.java
```diff
@@ -406,9 +407,12 @@ private ServiceFunctionChain validateAndLoadSfcBind(EntityManager em, BindSecuri
...
              
              //if this virtual system is pointing to more than one SFC and one of the SFC other than given SFC
-             //is already binded to SG and active, throw exception
-            if (vs.getServiceFunctionChains().size() > 1) {
-                for (ServiceFunctionChain serviceFunctionChain : vs.getServiceFunctionChains()) {
+             //is already binded to SG and active, throw exception;
+             //Get All the sfcs having this virtual system in the chain
+             List<ServiceFunctionChain> serviceFunctionChainList = ServiceFunctionChainEntityMgr.
+                                                                 listServiceFunctionChainsByVirtualSystem(em, vs);
+            if (serviceFunctionChainList != null) {
```
arvindn05: dont need null check

osc-server/src/main/java/org/osc/core/broker/service/securitygroup/BindSecurityGroupService.java
```diff
@@ -440,7 +444,7 @@ private void checkVirtualSystemRedundancyInServiceFunctionChains(EntityManager e
         List<SecurityGroup> sgList = SecurityGroupEntityMgr.listSecurityGroupsBySfcIdAndProjectId(em, sfcId,
```
arvindn05: i think larkins has a query to list other SG's by id and project id....you can probably create this method the same way
see
https://github.com/opensecuritycontroller/osc-core/pull/517/files#diff-cb0bcdaa1792493a7aceb46f5207ce59R129

osc-server/src/main/java/org/osc/core/broker/service/securitygroup/BindSecurityGroupService.java
karimull: This not visible in unit test case to mock ApifactoryServices. Hence changed to public.

osc-server/src/main/java/org/osc/core/broker/service/securitygroup/BindSecurityGroupService.java
karimull: Yes we will take off any querying redundant with integration test...For now we will have it.

osc-server/src/main/java/org/osc/core/broker/service/securitygroup/BindSecurityGroupService.java
karimull: removed

osc-server/src/main/java/org/osc/core/broker/service/persistence/ServiceFunctionChainEntityMgr.java
karimull: For the first comment.  I found it during unit test case and I opted  for querying than getting from vs.sfc list here is explanation for the bug

2.When checking any existing SFC other than the given SFC to bind connected to a SG and particular Tenant(Redundancy chain check), was getting SFC list from Virtual Systems. But in Add or update of SFC entity services we were not updating SFC list in VirtualSystem Entity. So added a query instead of getting from virtualsystem sfc entity list, which is more preferred.


For the Second comment.. sorting by name...nothing in particular..was just going with example code.. will see if I can remove it.

karimull: removed sorting by name...

PR #520, opened 2017-10-07
 add unit tests for security group member hook task 

arvindn05:


osc-server/src/test/java/org/osc/core/broker/service/tasks/conformance/openstack/securitygroup/SecurityGroupMemberHookCheckTaskTestData.java
```diff
@@ -0,0 +1,212 @@
...
+        } catch (Exception e) {
+            throw new RuntimeException();
+        }
+        for (VMPort port : sgm.getVmPorts()) {
+            expectedGraph.appendTask(new SfcFlowClassifierCreateTask().create(sgm.getSecurityGroup(), port));
+        }
+        return expectedGraph;
+    }
+
+    public TaskGraph getSfcWithDeletedPortsTaskGraph(SecurityGroupMember sgm) {
```
emanoelxavier: can these graph methods be static?

arvindn05: Thats what i tried initially but the api factory and sdn api being shared was causing interference and test were failing

emanoelxavier: I see, can you put a TODO on that so we can take a look later? Maybe something we can improve also on the overall test approach.

arvindn05: will add it as part of next commit

PR #525, opened 2017-10-09
Enabling delete DS for k8s

emanoelxavier:
1. Updating tasks to include k8s DS deletion.
2. Enforcing that DS, DA and VS deletion cannot happen if the ports are assigned/allocated to a protected port.


osc-server/src/main/java/org/osc/core/broker/service/validator/DistributedApplianceDtoValidator.java
```diff
@@ -139,6 +142,7 @@ void validate(DistributedApplianceDto dto, boolean forCreate) throws Exception {
                     "The associated Appliance Manager Connector must be selected for this Distributed Appliance.");
         }
 
+        List<Long> existingVsIds = new ArrayList<>();
```
dmitryintel: "existing" Some of  these could be newly created.

PR #526, opened 2017-10-09
Set poms to runtime. Add LogUtil.

dmitryintel:
* Some pom dependencies on logback-classic were redundant. Others could be left as runtime scope.
* Add back logging System.out  PrintStream: some initial log entries were lost without it.
* ServerControl can now easily use logback mechanism instead System.out.

### DIFF
--- a/osc-control/src/main/java/org/osc/core/server/control/ServerControl.java
+++ b/osc-control/src/main/java/org/osc/core/server/control/ServerControl.java
@@ -110,8 +110,8 @@ public class ServerControl {
             } else if (cmd.hasOption("check")) {
                 // --check is used by vmidc.sh
                 if (isRunningServer()) {
-                    System.out.println(ServerControl.PRODUCT_NAME + " Server is already running.");
-                    System.out.println(getServerVersion());
+                    log.info(ServerControl.PRODUCT_NAME + " Server is already running.");
+                    log.info(getServerVersion());
                     exitStatus = 2; // already running
                 }
             } else if (cmd.hasOption("version")) {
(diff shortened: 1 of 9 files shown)


osc-server/pom.xml
```diff
@@ -165,13 +165,6 @@
             <version>${slf4j.version}</version>
             <scope>provided</scope>
         </dependency>
-         <dependency>
```
emanoelxavier: It is strange that this is not needed here but it is needed in the control project. Should it be the other way around?

osc-control/src/main/java/org/osc/core/server/control/ServerControl.java
```diff
@@ -110,8 +110,8 @@ public boolean isRunning() {
             } else if (cmd.hasOption("check")) {
                 // --check is used by vmidc.sh
                 if (isRunningServer()) {
-                    System.out.println(ServerControl.PRODUCT_NAME + " Server is already running.");
-                    System.out.println(getServerVersion());
+                    log.info(ServerControl.PRODUCT_NAME + " Server is already running.");
```
arvindn05: this should be sysout. If the server is running, then server control will not launch and its log is not setup...

osc-server/src/main/java/org/osc/core/broker/util/log/LogUtil.java
```diff
@@ -0,0 +1,71 @@
...
+
+import java.io.PrintStream;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class LogUtil {
+
+    public static void initLog() {
```
arvindn05: rename to redirectConsoleMessagesToLog

osc-server/pom.xml
dmitryintel: I actually observed it to fail without these dependencies. But with my recent change these dependencies are not needed.

PR #527, opened 2017-10-09
 Added code to validate DS delete  for SFC, do not allow delete if DS…

karimull:
… virtual system is attached to a SFC


karimull:
Pushed changes Emanoel mentioned and did following test
1. Delete DA and DS with VC-type neutron Sfc, and sfc attached to the Virtual system: results as expected.(error - msg)
2.  Delete DA and DS with VC-type neutron sfc : Results as expected (got deleted).
3. Delete DA and DS with VC-type non-neutronSfc : Results as expected ( was able to delete)

PR #534, opened 2017-10-10
Fixing LazyInitialization issue

larkinscarvalho:


osc-server/src/test/java/org/osc/core/broker/service/tasks/conformance/openstack/sfc/SfcFlowClassifierDeleteTaskTest.java
```diff
@@ -115,6 +115,8 @@ public void testInitialize() throws Exception {
         when(this.port.getVm()).thenReturn(mockVm);
         when(this.port.getInspectionHookId()).thenReturn(INSPECTION_HOOK_ID);
 
+        Mockito.when(this.em.find(VMPort.class, this.port.getId())).thenReturn(this.port);
+
```
emanoelxavier: extra ln

osc-server/src/test/java/org/osc/core/broker/service/tasks/conformance/openstack/sfc/SfcFlowClassifierUpdateTaskTest.java
```diff
@@ -121,6 +121,8 @@ public void testInitialize() throws Exception {
         when(this.port.getVm()).thenReturn(mockVm);
         when(this.port.getInspectionHookId()).thenReturn(INSPECTION_HOOK_ID);
 
+        Mockito.when(this.em.find(VMPort.class, this.port.getId())).thenReturn(this.port);
+
```
emanoelxavier: extra ln

osc-server/src/test/java/org/osc/core/broker/service/tasks/conformance/openstack/sfc/SfcFlowClassifierDeleteTaskTest.java
emanoelxavier: also on ln 112 since this is mocked i think getId will always return null, you may want to mock getId from the vmport and make it return something else other than null.

PR #535, opened 2017-10-11
Fix the log size to 10 MB.

dmitryintel:
100MB / 10 files == 10 MB / file


dmitryintel:
100MB / 10 files == 10 MB / file

PR #537, opened 2017-10-11
 Remove SFC link on SG when force deleted

karimull:


osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/securitygroup/SecurityGroupUpdateOrDeleteMetaTask.java
```diff
@@ -241,6 +241,11 @@ private void buildTaskGraph(EntityManager em, boolean isDeleteTg, String domainI
         VmDiscoveryCache vdc = new VmDiscoveryCache(this.sg.getVirtualizationConnector(),
                 this.sg.getVirtualizationConnector().getProviderAdminProjectName());
 
+        //remove sfc link to SG if this sg is either marked for deletion or isDeleteTg is set
+        if (this.sg.getMarkedForDeletion() || isDeleteTg) {
+            this.sg.setServiceFunctionChain(null);
```
arvindn05: it should be set as part of delete service right... not in the job

osc-server/src/main/java/org/osc/core/broker/service/securitygroup/DeleteSecurityGroupService.java
```diff
@@ -77,6 +78,7 @@ public BaseJobResponse exec(BaseDeleteRequest request, EntityManager em) throws
 
                 // Mark this security Group for Deletion and Trigger  Sync Job
                 OSCEntityManager.markDeleted(em, securityGroup, this.txBroadcastUtil);
+                //remove sfc link to SG
```
arvindn05: comment needs to move to line 67?

osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/securitygroup/ForceDeleteSecurityGroupTask.java
```diff
@@ -31,9 +31,9 @@
 import org.osc.core.broker.service.persistence.OSCEntityManager;
 import org.osc.core.broker.service.persistence.SecurityGroupEntityMgr;
 import org.osc.core.broker.service.tasks.TransactionalTask;
-import org.slf4j.LoggerFactory;
 import org.osgi.service.component.annotations.Component;
 import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
arvindn05: revert file if no changes

osc-server/src/main/java/org/osc/core/broker/service/securitygroup/DeleteSecurityGroupService.java
```diff
@@ -64,6 +64,9 @@ public BaseJobResponse exec(BaseDeleteRequest request, EntityManager em) throws
         try {
             unlockTask = LockUtil.tryLockSecurityGroup(securityGroup,
                     securityGroup.getVirtualizationConnector());
+            //remove sfc link to SG
+            securityGroup.setServiceFunctionChain(null);
```
arvindn05: this has no effect on the force delete since this transaction is not committed before the job starts...

Assuming force delete happens after the regular delete, it might be fine.

Issue #540, opened 2017-10-11
Do not allow binding in neutron-sfc plugin from UI


karimull:
### Description
SFC binding is only allowed through APIs as UI Is not supported yet.

### Expected Behavior
User should shown an error when trying to bind services to SG supporting neutron-sfc .
### Actual Behavior
SFC binding is allowed through UI.
### Steps to Reproduce
create VirtualizationConnector- with neutron sfc
create Security group
Bind a Security Group  with services.
### Additional Information
Arvind found this issue during story acceptance . Tracing as an issue
### Environment
- OSC Version/Build:  108
- Cloud Provider Info:
- Plugin Info:
- Other:

### Status
- [x ] Reproduced
- [x ] In Progress
- [ ] Validated     =>    Verified in Version/Build:

arvindn05:
@karimull just to clarify, the issue is not the below that SFC binding is allowed via UI

In neutron SFC case, creating individual Security Group interfaces via the bind dialog/service should not be allowed. Currently this is allowed.

arvindn05:
DE5294

mrajora-1:
Verified with Version: 1.0.0-SNAPSHOT (Build:123-gbf90939, 2017-10-18T07:20:37Z)
Now I am getting an error when binding or unbinding SFC-SG from UI

![bug-540-no-binding-from-ui](https://user-images.githubusercontent.com/28600997/31739449-c302d556-b402-11e7-8806-cc5d2472fb79.JPG)

![bug-540-no-unbinding-from-ui](https://user-images.githubusercontent.com/28600997/31739493-eb06b1d0-b402-11e7-8dbb-6038ec5feea2.JPG)

mrajora-1:
Note that the error is same whether we bind or unbind. Was thinking that unbind error should have been different. Discussed with Karim and he suggested that the same error was kept as the operation is still bind.

PR #552, opened 2017-10-13
 Issue #540 : User should not be allowed to bind an SG from OSC UI wh…

karimull:
…en there is an SFC configured from the API(DE5294)

### DIFF
--- a/osc-server/src/main/java/org/osc/core/broker/service/securitygroup/BindSecurityGroupService.java
+++ b/osc-server/src/main/java/org/osc/core/broker/service/securitygroup/BindSecurityGroupService.java
@@ -361,7 +361,7 @@ public class BindSecurityGroupService extends ServiceDispatcher<BindSecurityGrou
             //user is trying to configure binding SFC with either UI which currently not supported or APIs that are not
             //valid for SFC binding, so throw an exception.
             throw new VmidcBrokerValidationException(String.format(
-                    "Unsupported Interface or API to bind/unbind Security Group %s to Virtual System with a Service Function Chain",
+                    "Binding individual Virtual Systems/Services is not supported for the Security Group %s. Please bind to a Service Function Chain",
                     this.securityGroup.getName()));
         }
 		if(request.getSfcId() == null) {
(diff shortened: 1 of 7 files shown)


osc-server/src/main/java/org/osc/core/broker/service/securitygroup/BindSecurityGroupService.java
```diff
@@ -368,6 +357,13 @@ private VirtualSystem validateAndLoadVirtualSystem(EntityManager em, Long vsId)
 	private ServiceFunctionChain validateAndLoadSfcBind(EntityManager em, BindSecurityGroupRequest request) throws Exception {
 
     	ServiceFunctionChain sfc = null;
+        if (!request.isBindSfc()) {
+            //user is trying to configure binding SFC with either UI which currently not supported or APIs that are not
+            //valid for SFC binding, so throw an exception.
+            throw new VmidcBrokerValidationException(String.format(
+                    "Unsupported Interface or API to bind/unbind Security Group %s to Virtual System with a Service Function Chain",
```
arvindn05: since this is called via the UI as well. We should just say "Binding individual Virtual Systems/Services is not supported for the Security Group %s. Please bind to a Service Function Chain"

osc-server/src/test/java/org/osc/core/broker/service/BindSecurityGroupServiceTest.java
```diff
@@ -465,14 +467,15 @@ public void testValidate_WhenSfcVsDoNotMatchWithBindSeviceVs_ThrowsValidationExc
...
                 .thenReturn(VALID_SG.getVirtualizationConnector());
         Mockito.when(this.em.find(Mockito.eq(ServiceFunctionChain.class), Mockito.eq(VALID_ID))).thenReturn(sfc);
         Mockito.when(this.bindSecurityGroupService.apiFactoryService.supportsNeutronSFC(VALID_SG)).thenReturn(true);
         this.exception.expect(VmidcBrokerValidationException.class);
-        this.exception.expectMessage("Binding request Virutal System Id : " + INVALID_ID
-                + " do not match with any of the id in Service Function Chain's Virtual System Id list");
+        this.exception.expectMessage("Binding request Virtual System Id : " + serviceToBindTo.getVirtualSystemId() +
+                " do not match with any of the ids in Service Function Chain Virtual System Id list");
 
         // Act
```
arvindn05: can you add a unit test for the case you encountered now?

Issue #539, opened 2017-10-11
Upon fresh ovf install RestClient errors on screen and in boot.log


dmitryintel:
### Description

Upon fresh ovf install, multiple RestClient exceptions show up in **/var/log/boot.log** and on the Console screen. 

### Expected Behavior

The exceptions should go to **/opt/vmidc/log/securityBorker.log** (if they happen) and should not be seen in the console or in **boot.log**.

### Actual Behavior

See the description.

### Steps to Reproduce

Install an ovf and watch the console in VSphere.

### Additional Information

The exceptions themselves may well have been happening before Oct 6, 2017, even with commits  at or before [3b67a0ee](https://github.com/opensecuritycontroller/osc-core/commit/3b67a0ee160bf7f599e617e3bd7c0ea01215e560). But with our slf4j logging changes they started appearing on the console. 

Typical exception:

18:54:21.415 [main] DEBUG org.osc.core.broker.rest.client.RestBaseClient - Error invoking getResources
org.osc.core.broker.service.exceptions.RestClientException: Failed to GET resource: 'https://localhost:8090/api/server/v1/serverManagement/status'.
Reason: 'javax.ws.rs.ProcessingException: java.net.ConnectException: Connection refused (Connection refused)'.
Response Code: null.
Response: 'null'.
        at org.osc.core.broker.rest.client.RestBaseClient.createRestClientException(RestBaseClient.java:504)
        at org.osc.core.broker.rest.client.RestBaseClient.getResource(RestBaseClient.java:269)
        at org.osc.core.broker.rest.client.RestBaseClient.getResource(RestBaseClient.java:246)
        at org.osc.core.broker.rest.client.RestBaseClient.getResource(RestBaseClient.java:241)
        at org.osc.core.server.control.ServerControl.getServerStatusResponse(ServerControl.java:356)
        at org.osc.core.server.control.ServerControl.isRunningServer(ServerControl.java:279)
        at org.osc.core.server.control.ServerControl.access$000(ServerControl.java:52)
        at org.osc.core.server.control.ServerControl$1.isRunning(ServerControl.java:102)
        at org.osc.core.broker.util.ServerUtil.startServerProcess(ServerUtil.java:290)
        at org.osc.core.broker.util.ServerUtil.startServerProcess(ServerUtil.java:274)
        at org.osc.core.server.control.ServerControl.main(ServerControl.java:99)
[...]

PR #554, opened 2017-10-13
Copy the logback config so startup exceptions go to the correct log.

dmitryintel:
* See issue [#539](https://github.com/opensecuritycontroller/osc-core/issues/539)

* ServerControl is reverted to almost exactly what it was before logging changes.


osc-control/src/main/java/org/osc/core/server/control/ServerControl.java
```diff
@@ -48,9 +48,14 @@
 import org.osc.core.broker.util.ServerUtil.ServerServiceChecker;
 import org.osc.core.broker.util.VersionUtil;
 import org.osc.core.broker.util.db.DBConnectionParameters;
+import org.osc.core.broker.util.log.LogUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class ServerControl {
 
+    static Logger log = LoggerFactory.getLogger(ServerControl.class);
```
arvindn05: can be private?

PR #555, opened 2017-10-16
Deleting the manager SG if the SGI is deleted

### LINKED ISSUE #528: Unbinding and binding the same SG for k8s is failing
### Description
Unbinding and re-biding a Kubernetes is failing due to redundant create of a security group in the appliance manager.

### Expected Behavior
Unbinding and re-binding a k8s security group should work consistently.

### Actual Behavior
Unbinding and re-binding a k8s security group triggers a synchronization job which is failing with the message "Security group with name xxx already exists"

### Steps to Reproduce
1. Create a SG, DA and DS for Kubernetes using the OSC REST Apis.
2. Bind the SG to the VS using the OSC REST Apis.
3. Unbind the SG from the VS using the OSC UI (or REST APIs).
4. Bind the SG to the VS using the OSC REST Apis.
5. Observer the job failure.

### Additional Information
image

### Workaround
Force delete the unbound security group and create a new one with a different name and same member labels. The new one can then be successfully bound.

### Environment
- OSC Version/Build: Build:100-g2bfbb51 
- Cloud Provider Info: K8s v1.6
- Plugin Info: ISM and NSC plugin.
- Other:

### Status
- [x] Reproduced
- [x] In Progress

emanoelxavier:
Fixes #528

### DIFF
--- a/osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/securitygroup/MgrSecurityGroupCheckMetaTask.java
+++ b/osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/securitygroup/MgrSecurityGroupCheckMetaTask.java
@@ -30,168 +30,170 @@ import org.osc.core.broker.model.plugin.ApiFactoryService;
 import org.osc.core.broker.service.persistence.SecurityGroupEntityMgr;
 import org.osc.core.broker.service.persistence.SecurityGroupInterfaceEntityMgr;
 import org.osc.core.broker.service.tasks.TransactionalMetaTask;
-import org.slf4j.LoggerFactory;
 import org.osc.sdk.manager.api.ManagerSecurityGroupApi;
 import org.osc.sdk.manager.element.ManagerSecurityGroupElement;
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Reference;
 import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 @Component(service = MgrSecurityGroupCheckMetaTask.class)
 public class MgrSecurityGroupCheckMetaTask extends TransactionalMetaTask {
-	private static final Logger log = LoggerFactory.getLogger(MgrSecurityGroupCheckMetaTask.class);
+    private static final Logger log = LoggerFactory.getLogger(MgrSecurityGroupCheckMetaTask.class);
 
-	@Reference
-	private CreateMgrSecurityGroupTask createMgrSecurityGroupTask;
+    @Reference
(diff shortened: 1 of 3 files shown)


osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/securitygroup/MgrSecurityGroupCheckMetaTask.java
```diff
@@ -30,168 +30,170 @@
...
+        } else {
+            // Sync members only for the specified security group
+            // For a given VS and SG, we will only have one SGI
+            SecurityGroupInterface sgi = SecurityGroupInterfaceEntityMgr
+                    .findSecurityGroupInterfacesByVsAndSecurityGroup(em, this.vs, this.sg);
+            ManagerSecurityGroupElement mepg = mgrSgApi.getSecurityGroupById(sgi.getMgrSecurityGroupId());
+            if (mepg == null) {
+                // Add new security group to Manager
+                this.tg.appendTask(this.createMgrSecurityGroupTask.create(this.vs, this.sg, sgi));
+            } else if (sgi.getMarkedForDeletion() == null || !sgi.getMarkedForDeletion()){
```
emanoelxavier: This condition is the only meaningful change on this metatask.

arvindn05: dont need null check. Marked for deletion should always have a default value of false as per the base entity.

osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/securitygroup/MgrSecurityGroupCheckMetaTask.java
```diff
@@ -30,168 +30,170 @@
...
+                    this.tg.appendTask(this.deleteMgrSecurityGroupTask.create(this.vs, mepge));
+                }
+            }
+        } else {
+            // Sync members only for the specified security group
+            // For a given VS and SG, we will only have one SGI
+            SecurityGroupInterface sgi = SecurityGroupInterfaceEntityMgr
+                    .findSecurityGroupInterfacesByVsAndSecurityGroup(em, this.vs, this.sg);
+            ManagerSecurityGroupElement mepg = mgrSgApi.getSecurityGroupById(sgi.getMgrSecurityGroupId());
+            if (mepg == null) {
```
arvindn05: you should add the same check here. Only if SGI is not marked for deletion the create/update actions are valid right?

Issue #394, opened 2017-07-27
api-doc MC GET and POST have wrong name "PolicyMappingSupported", should be "isPolicyMappingSupported"


alin2k:
### Description
In https://OSCipaddress:8090/api-doc, Operations for Manager Connectors section there are three GETs, POST, PUT all have field name "PolicyMappingSupported", POST did get "bad request, un-recognize field", so all 5 places need to name it as "**is**PolicyMappingSupported". POST may even take this field out (not needed)
### Expected Behavior
User can use api-doc to test Rest API functions OSC provides
### Actual Behavior
POST/PUT failed. GET Model Schema display can confuse user with wrong name.

### Steps to Reproduce
1)https://OSCipaddress:8090/api-doc, Operations for Manager Connectors section
2)GET Model Schema display in the yellow box
3)POST/PUT click write box and real fields in yellow box has wrong name and causes request fauiled
### Additional Information
screenshot attached.
### Environment
- OSC Version/Build: v1B30
- Cloud Provider Info:
- Plugin Info:
- Other:

### Status
- [x] Reproduced
- [x] In Progress
- [ ] Validated     =>    Verified in Version/Build: 

<img width="413" alt="apidocgetmcs" src="https://user-images.githubusercontent.com/11319375/28649308-d4987b16-7228-11e7-91a6-487da586bd24.PNG">
<img width="386" alt="apidocpostmcs" src="https://user-images.githubusercontent.com/11319375/28649309-d9b4109c-7228-11e7-9054-304ef414af5c.PNG">

alin2k:
v0.8 Build#118: Looks like fixed doc in MC GET but not in MC POST? (Issue is still OPEN):
...
"policyMappingSupported": false

larkinscarvalho:
@alin2k How did you verify the MC GET? 
Issue is not yet fixed. Also, the dev task is in progress on rally.

alin2k:
- [x] Validated     =>    Verified fix in Version/Build: v0.8 Build#123

PR #560, opened 2017-10-16
Fix API doc parameters

larkinscarvalho:
Swagger creates the model and document based on the getters. 

@ApiModelProperty(name = "testName") - seems to have no effect as per the design https://github.com/springfox/springfox/issues/1289
https://github.com/swagger-api/swagger-core/issues/2134

We can use @XmlElement(name = "testName") but that will just change the response provided and not  the actual documentation.

Updating getters seems to be the only solution

### DIFF
--- a/osc-service-api/src/main/java/org/osc/core/broker/service/dto/ApplianceManagerConnectorDto.java
+++ b/osc-service-api/src/main/java/org/osc/core/broker/service/dto/ApplianceManagerConnectorDto.java
@@ -55,7 +55,7 @@ public class ApplianceManagerConnectorDto extends BaseDto {
     private Long lastJobId;
 
     @ApiModelProperty(readOnly = true, value = "Determines whether the appliance manager supports policy mapping.")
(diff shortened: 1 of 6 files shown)


larkinscarvalho:
Fixes https://github.com/opensecuritycontroller/osc-core/issues/394

osc-service-api/src/main/java/org/osc/core/broker/service/dto/ApplianceManagerConnectorDto.java
```diff
@@ -135,7 +135,7 @@ public void setLastJobId(Long lastJobId) {
         this.lastJobId = lastJobId;
     }
 
-    public Boolean isPolicyMappingSupported() {
+    public Boolean getIsPolicyMappingSupported() {
```
arvindn05: line comment to indicate why the non-standard naming, just in case someone comes in and "fixes" this and causes regressions

larkinscarvalho:
@arvindn05 Tested locally using API's

PR #563, opened 2017-10-17
Fix VC API issue where rabbitmq password is not accepted unless its encrypted

### LINKED ISSUE #445: OSC Add VC thru UI works but using API got failed decrypt cipher AES-CTR
### Description
Automation used API to Add VC and Job failed, got failed to decrypt cipher AES-CTR
### Expected Behavior
Add VC as UI the same connection with Job passed
### Actual Behavior
Using API to Add VC Job failed
### Steps to Reproduce
Using Python Automation DA.robot Curl sample:
curl -X POST --header 'Content-Type: application/xml' --header 'Accept: application/xml' -d '{
  "id": 0,
  "parentId": 0,
  "name": "vc2",
  "type": "OPENSTACK",
  "controllerIP": "",
  "controllerUser": "",
  "controllerPassword": null,
  "providerIP": "10.3.241.154",
  "providerUser": "admin",
  "providerPassword": "admin123",
  "softwareVersion": "Icehouse",
  "controllerType": "NSC",
  "providerAttributes": {},
  "adminProjectName": "admin",
  "adminDomainId": "default",
  "lastJobState": "COMPLETED",
  "lastJobStatus": "PASSED",
  "lastJobId": 0,
  "skipRemoteValidation": true,
  "forceAddSSLCertificates": false
}' 'https://10.3.240.156:8090/api/server/v1/virtualizationConnectors'
### Environment
- OSC Version/Build: v1b49
- Cloud Provider Info:
- Plugin Info:
- Other:

### Status
- [ ] Reproduced
- [ ] In Progress

[APiAddVc-aesctrFailLog.txt](https://github.com/opensecuritycontroller/osc-core/files/1235662/APiAddVc-aesctrFailLog.txt)

arvindn05:
Fixes #445

### DIFF
--- a/osc-server/src/main/java/org/osc/core/broker/service/persistence/VirtualizationConnectorEntityMgr.java
+++ b/osc-server/src/main/java/org/osc/core/broker/service/persistence/VirtualizationConnectorEntityMgr.java
@@ -16,7 +16,9 @@
  *******************************************************************************/
 package org.osc.core.broker.service.persistence;
 
+
 import static java.util.stream.Collectors.*;
+import static org.osc.core.common.virtualization.VirtualizationConnectorProperties.*;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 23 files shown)


osc-server/src/main/java/org/osc/core/broker/service/persistence/VirtualizationConnectorEntityMgr.java
```diff
@@ -101,6 +109,12 @@ public static void fromEntity(VirtualizationConnector vc, VirtualizationConnecto
         dto.setAdminProjectName(vc.getProviderAdminProjectName());
         dto.setAdminDomainId(vc.getAdminDomainId());
         dto.getProviderAttributes().putAll(vc.getProviderAttributes());
+
+        // For rabbit MQ password, decrypt it before setting it on the dto.
+        String rabbitMqPassword = dto.getProviderAttributes().get(ATTRIBUTE_KEY_RABBITMQ_USER_PASSWORD);
```
emanoelxavier: This will not be exposed through the rest api right?

arvindn05: yes, it will not be exposed in the API.
The following call as part of the list and get services will remove the password from the dto
org.osc.core.broker.service.dto.VirtualizationConnectorDto.sanitizeVirtualizationConnector(VirtualizationConnectorDto)

Issue #549, opened 2017-10-13
When binding same SFC to more than one SG, the service names of virtual system names are not in order


karimull:
### Description
When binding same SFC to more than one SG, the service names of virtual system names are not in order for the second SG
### Expected Behavior
For both SG the services names should be in order
### Actual Behavior
Second SG service names are not in order
### Steps to Reproduce
1. Create VC with n-sfc plugin
2. create SFC with at least two virtual systems
3. Bind this SFC with more than one SG.
4. Should see the problem in SG screen under "Services"
### Additional Information
Arvind, saw this issue during Demo. Opening bug on his behalf
### Environment
- OSC Version/Build: Build 112
- Cloud Provider Info:
- Plugin Info:
neutron-sfc
- Other:

### Status
- [ ] Reproduced
- [x] In Progress
- [ ] Validated     =>    Verified in Version/Build:

mrajora-1:
Verified with - Version: 1.0.0-SNAPSHOT (Build:140-gb301029, 2017-10-24T07:22:46Z)
Order of services looks fine with build 140 

![da-order-2-1-w-sfc-bind-ok](https://user-images.githubusercontent.com/28600997/32013080-cc77c4fe-b96e-11e7-9d26-b19f4606c058.JPG)

![da-order-w-sfc-bind-ok](https://user-images.githubusercontent.com/28600997/32013097-d7adb7f2-b96e-11e7-98ad-527e9eab7ce3.JPG)

[securityBroker-da-order-ok.log](https://github.com/opensecuritycontroller/osc-core/files/1415533/securityBroker-da-order-ok.log)

PR #564, opened 2017-10-17
Issue#549 , Even though the order is not an input for Virtual systems…

karimull:
… services, add order to show the service list internally


osc-server/src/main/java/org/osc/core/broker/service/securitygroup/BindSecurityGroupService.java
```diff
@@ -108,12 +108,13 @@ public BaseJobResponse exec(BindSecurityGroupRequest request, EntityManager em)
                     }
```
emanoelxavier: Can you fix this indentation? This block is not correctly indented.

arvindn05:
fixes #549

Issue #524, opened 2017-10-09
Unable to delete the plugin from plugin menu option


sudhirappaji:
### Description
After installing the OSC from Build#96, delete option does not appear due to nauge plugin OSGI error. 
### Actual Behavior
deletion button on the plugin should appear.
### Steps to Reproduce
- Deploy the OSC Build#96.
- Go to Plugin Menu
- Delete option for the plugin does not appear due to nauge plugin error message is too long.
### Environment
- OSC Version/Build: OSC Build#95.
- Cloud Provider Info:
- Plugin Info:
- Other:

### Status
- [x] Reproduced
- [x] In Progress
- [ ] Validated     =>    Verified in Version/Build:

alin2k:
- [x ] Validated     =>    Verified fix in Version/Build: v0.8 Build#123

emanoelxavier:
This issue is still happening on later versions of OSC. For this to repro though one needs a defective plugin.

alin2k:
- [x] Validated     =>    Verified fix in Version/Build: v0.8 Build#125

PR #567, opened 2017-10-18
Setting the max width for the plugin info column

emanoelxavier:


emanoelxavier:
Fixes #524

osc-ui/src/main/java/org/osc/core/broker/view/maintenance/PluginsLayout.java
```diff
@@ -100,15 +101,16 @@ public PluginsLayout(BundleContext ctx, ImportPluginServiceApi importPluginServi
         this.plugins.addContainerProperty(PROP_PLUGIN_INFO, String.class, null);
         this.plugins.addContainerProperty(PROP_PLUGIN_DELETE, Button.class, null);
 
+        this.plugins.setColumnWidth(PROP_PLUGIN_INFO, PLUGIN_INFO_COLUMN_WIDTH);
```
arvindn05: can we add a tooltip so we can look at the longer error? or if we can expand the column and scroll, that should be fine as well.

osc-ui/src/main/java/org/osc/core/broker/view/maintenance/PluginsLayout.java
```diff
@@ -100,15 +104,33 @@ public PluginsLayout(BundleContext ctx, ImportPluginServiceApi importPluginServi
...
+
+        // Add a tooltip to the error column so the user is able to see the
+        // complete error message
+        this.plugins.setItemDescriptionGenerator(new ItemDescriptionGenerator() {
```
arvindn05: you can use lamda expr.

I would recommend to enable the save action in java which automatically converts annoymous classes to lambdas.

PR #569, opened 2017-10-19
make sure to remove api from map after they are closed

### LINKED ISSUE #523: Creation of Deployment Specification fails on a fresly installed sdnnsc plugin
### Description
Failure while creating the DS, with the following error message for the JOB task  “Deploying SVA for Distributed Appliance Instance …”  with the error message “ The Resource Provided has been reclaimed because the factory service that provides it was released”.
### Expected Behavior
DS creation should be permitted and mandatory in the above scenario.
### Steps to Reproduce
DS failure with the message "The Resource Provided has been reclaimed because the factory service that provides it was released".
### Additional Information
With the following workaround scenario, the problem is not reproduced:
1.	Reload the sdnnsc plugin after deleting the existing DA/DS.
2.	Reset/reboot OSC
3.	Run OSC and create VC, MC,DA,DS, SG, SGBind, SGunBind, and delete SG, DS,DA,MC and VC.
4.	All of the above tests are passing with this work around.

### Environment
- OSC Version/Build: Version: 1.0.0-SNAPSHOT (Build:97-g601eb9e, 2017-10-07T19:30:12Z
- Cloud Provider Info:
- Plugin Info:
- Other:

### Status
- [ ] Reproduced
- [ ] In Progress

arvindn05:
fix #523

### DIFF
--- a/osc-server/src/main/java/org/osc/core/broker/model/plugin/ApiFactoryServiceImpl.java
+++ b/osc-server/src/main/java/org/osc/core/broker/model/plugin/ApiFactoryServiceImpl.java
@@ -16,7 +16,7 @@
  *******************************************************************************/
 package org.osc.core.broker.model.plugin;
 
-import static java.util.Collections.emptyList;
+import static java.util.Collections.*;
 import static org.osc.sdk.controller.Constants.*;
 import static org.osc.sdk.manager.Constants.*;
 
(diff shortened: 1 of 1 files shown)


osc-server/src/main/java/org/osc/core/broker/model/plugin/ApiFactoryServiceImpl.java
```diff
@@ -16,7 +16,7 @@
  *******************************************************************************/
 package org.osc.core.broker.model.plugin;
 
-import static java.util.Collections.emptyList;
+import static java.util.Collections.*;
```
dmitryintel: Is this star really necessary?

arvindn05: for static imports our threshold is 3. I think in my workspace its set to 1. Will reconfigure.

https://github.com/opensecuritycontroller/community/blob/master/development/eclipse.md#code-style

Issue #556, opened 2017-10-16
OSC released v0.6 saved database .dbb cannot be restored to v0.8 got CouldNotPrepareStatement


alin2k:
### Description
Attached saved DB .dbb file from version 0.6 which cannot be Restore to ovf installed v0.8, got error
### Expected Behavior
Previous released version0.6, Backup database should be able to Restore to the v0.8
### Actual Behavior
Got pop-up red error: Cannot prepare the statement (screenshot attached)
### Steps to Reproduce
1)Freshly install a new version 0.8 Build#110 which has no configuration
2)OSC-Manage-Server-Maintenence-RestoreDB the attached v06BrokerServerDBBackup_156.dbb which was Backup from a running v0.6 OSC
3)OSC-Setup-Virtulization Connectors (click the table) got popup error (was VC defined in v0.6 e.g.)
### Additional Information (attachements including Log):
![06to08restoredbfaileds](https://user-images.githubusercontent.com/11319375/31626281-f305f8ce-b25d-11e7-9973-bea571b38604.PNG)
[06TO08restoreDBfailedLog500.txt](https://github.com/opensecuritycontroller/osc-core/files/1388481/06TO08restoreDBfailedLog500.txt)
[060b12ServerSupportBundle.zip](https://github.com/opensecuritycontroller/osc-core/files/1388484/060b12ServerSupportBundle.zip)

### Environment
- OSC Version/Build: v0.8 Build#110
- Cloud Provider Info:
- Plugin Info:
- Other:

### Status
- [x] Reproduced
- [x] In Progress
- [ ] Validated     =>    Verified in Version/Build:

alin2k:
[x ] Validated     =>    Verified fix in Version/Build: - Build:135-gad10cd8, 2017-10-21T07:20:42Z

PR #573, opened 2017-10-20
Fix upgrade SQL syntax so 0.6 to 0.8 works. Issue556

dmitryintel:
This is to fix the [issue556](https://github.com/opensecuritycontroller/osc-core/issues/556) The upgrade never finishes so, after the upgrade a request for VC reults in referencing a nonexistent column. 

The upgrade code never made it to the ADD COLUMN statement.

### DIFF
--- a/osc-server/src/main/java/org/osc/core/broker/util/db/upgrade/ReleaseUpgradeMgr.java
+++ b/osc-server/src/main/java/org/osc/core/broker/util/db/upgrade/ReleaseUpgradeMgr.java
@@ -35,10 +35,10 @@ import org.osc.core.broker.service.api.server.EncryptionApi;
 import org.osc.core.broker.service.api.server.EncryptionException;
 import org.osc.core.broker.util.db.DBConnectionManager;
 import org.osc.core.broker.util.db.DBConnectionParameters;
-import org.slf4j.LoggerFactory;
 import org.osc.core.common.job.FreqType;
 import org.osc.core.common.job.ThresholdType;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


osc-server/src/main/java/org/osc/core/broker/util/db/upgrade/ReleaseUpgradeMgr.java
```diff
@@ -517,17 +517,22 @@ private static void upgrade80to81(Statement stmt) throws SQLException {
     }
 
     private static void upgrade79to80(Statement stmt) throws SQLException {
-        execSql(stmt, "alter table DISTRIBUTED_APPLIANCE_INSTANCE DROP COLUMN IF EXISTS nsx_agent_id, "
-                + "nsx_host_id, nsx_host_name, nsx_host_vsm_uuid, nsx_vm_id;");
+        execSql(stmt, "alter table DISTRIBUTED_APPLIANCE_INSTANCE DROP COLUMN IF EXISTS nsx_agent_id;"
```
arvindn05: alter table drop column should work with multiple columns  is it because of the if exists clause its not working?

http://www.h2database.com/html/grammar.html#alter_table_drop_column

Also we should follow the same pattern and run each sql statement under execSql

dmitryintel: Yes. Verified in SQL bench.

arvindn05: without the if exists clause does it work?

Issue #566, opened 2017-10-18
Receiving Error - Cannot find port group id, when editing DS and enabling/disabling hosts


mrajora-1:
### Description
Receiving Error - Cannot find port group id, when editing DS and enabling/disabling hosts
### Expected Behavior
- DS Sync should pass without errors while editing DS and enabling/disabling hosts
- Remove the exclamation after the port group ID in the error. 
### Actual Behavior
- Receiving Error - Cannot find port group id, when editing DS and enabling/disabling hosts
- There is also an exclamation at the end of the port ID
### Steps to Reproduce
- Create VC (SFC), MC, DA
- Create DS with 2 hosts selected - job is passed
- Edit the DS and disable host 2 - Job passed
- Edit the DS and enable host 1 and disable host 2 - Job fails with error - Cannot find port group by id: a3dfbf82-701d-424b-a51c-55b94a3de420!
- Edit the DA and add host 2 again - Still the DS Sync fails with the same error as above
### Additional Information
- DS is active on Openstack
- Port Group ID is seen on Get DS API
### Environment
- OSC Version/Build: Version: 1.0.0-SNAPSHOT (Build:118-g0a08063, 2017-10-17T07:21:29Z)
- Cloud Provider Info:
- Plugin Info:
- Other:

### Status
- [ ] Reproduced
- [ ] In Progress
- [ ] Validated     =>    Verified in Version/Build: 

![cannot-find-port-grp-by-id-error-editing-ds-w-2-hosts](https://user-images.githubusercontent.com/28600997/31701392-ea5b497c-b384-11e7-886a-0f260230e1cc.JPG)

![appliance-instance-active-on-openstack](https://user-images.githubusercontent.com/28600997/31701401-fd297bc8-b384-11e7-9399-a74084413097.JPG)

![port-group-id-seen-on-get-ds](https://user-images.githubusercontent.com/28600997/31701404-058a7010-b385-11e7-9539-40e9b82b3ba3.JPG)

[securityBroker.log](https://github.com/opensecuritycontroller/osc-core/files/1393465/securityBroker.log)

arvindn05:
DE5334

mrajora-1:
Verified with Version: 1.0.0-SNAPSHOT (Build:140-gb301029, 2017-10-24T07:22:46Z)
I edited the DS enabling/disabling host1 and host2. Each time the DS creation passed with no errors. 

[...]

PR #578, opened 2017-10-21
Fix the wrong portGroupId: Issue 566

dmitryintel:
* Make sure the svaAndDAIDelete tasks run after at least one daiAdd task.
* Tested with a two-hypervisor system.
* [issue 566](https://github.com/opensecuritycontroller/osc-core/issues/566)

### DIFF
--- a/osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/deploymentspec/DSUpdateOrDeleteMetaTask.java
+++ b/osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/deploymentspec/DSUpdateOrDeleteMetaTask.java
@@ -286,7 +286,7 @@ public class DSUpdateOrDeleteMetaTask extends TransactionalMetaTask {
 
         for (DistributedApplianceInstance dai : daisToDelete) {
             // Remove any extra sva/DAI
-            log.info("Removing DAI/SVA: " + dai.getName() + " for host: " + dai.getHostName());
+            log.info("Removing DAI/SVA: %s for host: %s", dai.getName(), dai.getHostName());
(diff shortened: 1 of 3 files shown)


osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/deploymentspec/DSUpdateOrDeleteMetaTask.java
```diff
@@ -286,6 +283,13 @@ private void conformToHostsSelection(Collection<String> selectedHosts,
                 }
             }
         }
+
+        for (DistributedApplianceInstance dai : daisToDelete) {
+            // Remove any extra sva/DAI
+            log.info("Removing DAI/SVA: " + dai.getName() + " for host: " + dai.getHostName());
```
arvindn05: nitpick but use slf4j parameterized logging. Going forward we want to use this patttern for logging

logger.info("Removing DAI/SVA: {} for host: {}", dai.getName(), dai.getHostName());

osc-server/src/test/java/org/osc/core/broker/service/tasks/conformance/openstack/deploymentspec/DSUpdateOrDeleteMetaTaskTestData.java
```diff
@@ -225,10 +225,12 @@ public static TaskGraph createDAIHostSelectedGraph(DeploymentSpec ds) {
 
     public static TaskGraph createDAIHostNotSelectedGraph(DeploymentSpec ds) {
```
arvindn05: can you add a unit test for this specific scenario? Basically have multiple  DAI's and make sure the tasks are created appropriately.

Issue #559, opened 2017-10-16
Logging into the OSC shell fails in some OSC builds


karimull:
### Description
This issue is observed in some intermittent builds. User is not able to login to OSC shell with given password.
### Expected Behavior
User should be able to login to OSC shell with given password.
### Actual Behavior
In couple of recent builds user was not able to login with given password. The login was rejected with invalid password
### Steps to Reproduce
OSC Build-13 is having this issue. 
### Additional Information
This issue is only observed in certain builds.
### Environment
- OSC Version/Build: 
Build1.0.0-SNAPSHOT-118-g0a08063 of OSC
- Cloud Provider Info:
- Plugin Info:
- Other:

### Status
- [ ] Reproduced
- [ ] In Progress
- [ ] Validated     =>    Verified in Version/Build:

arvindn05:
DE5335

alin2k:
[x ] Validated     =>    Verified fix in Version/Build: - Build:135-gad10cd8, 2017-10-21T07:20:42Z

PR #579, opened 2017-10-23
Save the salt in hexadecimal code rather than a string. When saving a…

karimull:
…s a string some of the control characters are not read correctly and thereby failed to login


arvindn05:
fixes #559

Issue #551, opened 2017-10-13
Disable SGI binding update for multi-policy mapping supported bindings


larkinscarvalho:
### Description
Multi-policy mapping is not supported through UI. Once, we create multi-policy mapping, we should not be able to update binding through UI.

### Expected Behavior
Disable SGI binding for existing multi-policy mapping

### Actual Behavior
Policy drop down menu is editable

### Steps to Reproduce

- Create VC (using NSC), MC, 1DA, 1DS, 1SG
- Bind SG with multiple policies through rest API
- Click on the SG Bind button to view the SGI binding
- Uncheck the service enabled checkbox
- Then, again check service enabled checkbox, Policy menu is editable

### Environment
- OSC Version/Build: 1.0.0-SNAPSHOT (Build:113)
- Cloud Provider Info:
- Plugin Info:
- Other:

### Status
- [x] Reproduced
- [x] In Progress
- [ ] Validated     =>    Verified in Version/Build: 

![sgi_policy_ui](https://user-images.githubusercontent.com/9063674/31565068-45900d00-b01a-11e7-83fb-a4c85cc16b20.JPG)

arvindn05:
@larkinscarvalho lets drop if its high effort or has high chance for regression. Please let us know your evaluation

arvindn05:
addressed by https://github.com/opensecuritycontroller/osc-core/pull/585

ychoma:
It works fine now:
Verfied with Version: 1.0.0-SNAPSHOT (Build:144-g9c773e4, 2017-10-27T07:21:25Z)

If have already multiple policies - the combo box for policies is disabled and if you uncheck and check and then click OK remembers the multiple policies.

If you uncheck and click OK  - it unbind

When checking it again - the combo box is enabled and you can choose a single policy that you want from the UI it works

When used SFC - the unbind is not allowed through UI as expected

PR #585, opened 2017-10-25
Allow policy update of SGI, in case if the manager is supporting multiple policies and only one policy is binded

larkinscarvalho:


arvindn05:
please add referencce to bug #551 
also update the bug with scenrios where the box is enabled/disabled

larkinscarvalho:
Fixes: https://github.com/opensecuritycontroller/osc-core/issues/551

If we have one policy bounded, we should be able to update the policies and the checkbox is enabled.

However, if we have a SGI with multiple policies bounded, we cannot update the policies and the checkbox is disabled.

arvindn05:
Just to confirm, unbinding the policy/SGI is allowed in all cases right?

osc-ui/src/main/java/org/osc/core/broker/view/vc/securitygroup/BindSecurityGroupWindow.java
```diff
@@ -418,9 +418,10 @@ private void serviceTableClicked(long itemId) {
 		}
 	}
 
-	private boolean isMultiplePoliciesSupported(Long vsId) {
+	private boolean isBindingWithMultiplePolicies(Long vsId) {
```
arvindn05: isBindedWithMultiplePolicies

osc-ui/src/main/java/org/osc/core/broker/view/vc/securitygroup/BindSecurityGroupWindow.java
```diff
@@ -418,9 +413,10 @@ private void serviceTableClicked(long itemId) {
 		}
 	}
 
-	private boolean isMultiplePoliciesSupported(Long vsId) {
+	private boolean isBindedWithMultiplePolicies(Long vsId) {
 		for (VirtualSystemPolicyBindingDto binding : this.allBindings) {
-			if (binding.getVirtualSystemId().equals(vsId) && binding.isMultiplePoliciesSupported()) {
+			if (binding.getVirtualSystemId().equals(vsId) && binding.isMultiplePoliciesSupported() && binding.isBinded()
```
arvindn05: can we extract the condition below into another method and use it in line 152 as well?

`binding.isMultiplePoliciesSupported() && binding.isBinded() && binding.getPolicyIds().size() > 1`

we can name the overloaded method the same and add javadoc to both methods.
`isBindedWithMultiplePolicies(VirtualSystemPolicyBindingDto binding)`

larkinscarvalho:
I tried it already.
But, to find and match the virtual system, becomes messy.

arvindn05:
@larkinscarvalho not sure i understand. you will have 2 methods called isBindedWithMultiplePolicies() should be straightforward..

your condition will become

`binding.getVirtualSystemId().equals(vsId) && isBindedWithMultiplePolicies(binding)`

osc-ui/src/main/java/org/osc/core/broker/view/vc/securitygroup/BindSecurityGroupWindow.java
```diff
@@ -418,9 +413,10 @@ private void serviceTableClicked(long itemId) {
 		}
 	}
 
-	private boolean isMultiplePoliciesSupported(Long vsId) {
+	private boolean isBindedWithMultiplePolicies(Long vsId) {
 		for (VirtualSystemPolicyBindingDto binding : this.allBindings) {
```
larkinscarvalho: @arvindn05 we iterate on **this.allBindings** to find the right binding. If we want to introduce change, we will have to update serviceTableClicked to find the right bind and provide to isBindedWithMultiplePolicies

larkinscarvalho: @arvindn05 we iterate on this.allBindings to find the right binding. If we want to introduce change, we will have to update serviceTableClicked to find the right bind and provide to isBindedWithMultiplePolicies

arvindn05: i dont see the issue...you will still iterate in the isBindedWithMultiplePolicies(Long vsId) method...so no changes should be needed.

larkinscarvalho:
@arvindn05 ready for CR. Thanks.

PR #587, opened 2017-10-26
Fixes #584 : Added validation for SFC binding before SFC update opera…

### LINKED ISSUE #584: SFC update operation must not be allowed, if given SFC is already binding to a Security Group
### Description
SFC update operation must not be allowed, if given SFC is already binding to a Security Group
### Expected Behavior
Sfc update operation should fail with validation expectation, if this sfc is already binding to a security group 
### Actual Behavior
SFC is updated with no error.
### Steps to Reproduce
1. Create virtualizationConnector with neutron-sfc
2. Create DA-1 and DS under this virtualizationConnector 
3. Create Security Group under this virtualizationConnector 
4. Create SFC with virtualsystem under DA-1
5. Bind SFC with the Security Group
6. Update SFC with either null list or a new virtual system
### Environment
- OSC Version/Build: 
osc build - 128
- Cloud Provider Info:
- Plugin Info:
- Other:

### Status
- [x ] Reproduced
- [x ] In Progress

karimull:
…tion.Now updated is only allowed if SFC is not binding to a security group

### DIFF
--- a/osc-server/src/main/java/org/osc/core/broker/service/validator/ServiceFunctionChainRequestValidator.java
+++ b/osc-server/src/main/java/org/osc/core/broker/service/validator/ServiceFunctionChainRequestValidator.java
@@ -98,7 +98,7 @@ public class ServiceFunctionChainRequestValidator
         String sgNames = sgList.stream().filter(sg -> !sg.getMarkedForDeletion()).map(sg -> sg.getName()).collect(Collectors.joining(", "));
         if (!sgNames.isEmpty()) {
             throw new VmidcBrokerValidationException(
-                    String.format("Cannot update Service Function Chain: %s, as it is binded to Security Group(s) %s",
+                    String.format("Cannot update Service Function Chain: '%s', as it is binded to Security Group(s) '%s'",
                             sfc.getName(), sgNames));
         }
 
--- a/osc-server/src/main/java/org/osc/core/broker/service/validator/ServiceFunctionChainRequestValidator.java
+++ b/osc-server/src/main/java/org/osc/core/broker/service/validator/ServiceFunctionChainRequestValidator.java
@@ -19,17 +19,20 @@ package org.osc.core.broker.service.validator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import javax.persistence.EntityManager;
 
 import org.apache.commons.collections4.CollectionUtils;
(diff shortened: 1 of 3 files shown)


osc-server/src/main/java/org/osc/core/broker/service/validator/ServiceFunctionChainRequestValidator.java
```diff
@@ -89,6 +92,16 @@ public ServiceFunctionChain validateAndLoad(AddOrUpdateServiceFunctionChainReque
 		}
 		ValidateUtil.checkMarkedForDeletion(sfc, sfc.getName());
 		validateCommon(request);
+
+        //Throw an exception if this sfc is already binded to a security group.
+        List<SecurityGroup> sgList = SecurityGroupEntityMgr.listSecurityGroupsBySfcId(this.em, sfc.getId());
+        String sgNames = sgList.stream().filter(sg -> !sg.getMarkedForDeletion()).map(sg -> sg.getName()).collect(Collectors.joining(", "));
+        if (!sgNames.isEmpty()) {
+            throw new VmidcBrokerValidationException(
+                    String.format("Cannot update Service Function Chain: %s, as it is binded to Security Group(s) %s",
```
arvindn05: use ' ' for the names.

osc-server/src/main/java/org/osc/core/broker/service/validator/ServiceFunctionChainRequestValidator.java
```diff
@@ -89,6 +92,16 @@ public ServiceFunctionChain validateAndLoad(AddOrUpdateServiceFunctionChainReque
 		}
 		ValidateUtil.checkMarkedForDeletion(sfc, sfc.getName());
 		validateCommon(request);
+
+        //Throw an exception if this sfc is already binded to a security group.
+        List<SecurityGroup> sgList = SecurityGroupEntityMgr.listSecurityGroupsBySfcId(this.em, sfc.getId());
+        String sgNames = sgList.stream().filter(sg -> !sg.getMarkedForDeletion()).map(sg -> sg.getName()).collect(Collectors.joining(", "));
+        if (!sgNames.isEmpty()) {
+            throw new VmidcBrokerValidationException(
+                    String.format("Cannot update Service Function Chain: '%s', as it is binded to Security Group(s) '%s'",
```
arvindn05: remove extra ,

> Cannot update Service Function Chain: '%s' as it is binded to Security Group(s) '%s'

Issue #550, opened 2017-10-13
Job Status first shows 'ABORTED' when binding a second SG with 2 DAs, then passes on subsequent Sync


mrajora-1:
### Description
Job Status first shows 'ABORTED' when binding a second SG with 2 DAs, then passes on subsequent Sync
### Expected Behavior
Jobs should have passed on the first attempt
### Actual Behavior
Job Status first shows 'ABORTED' when binding a second SG with 2 DAs, then passes on subsequent Sync
### Steps to Reproduce
- Create VC (using sfc), MC, 2DAs, 2DSs, 2SGs
- SG-1 is binded with da-1, da-2 (using SFC bind API)
- Bind SG-2 also with da-1, da-2 also  using SFC bind API
- The second bind shows an ABORTED job status. 
- Subsequent sync makes the job pass
### Additional Information
Excerpt from logs: 
2017-10-13 07:49:16,874 [ERROR| isc-job-pool-7| o.o.c.b.j.Job]: Fatal error during job execution (Job [name=Syncing Security Group 'sg-2', state=RUNNING, status=PASSED])
java.util.ConcurrentModificationException: null
	at java.util.HashMap$HashIterator.nextNode(HashMap.java:1437)
	at java.util.HashMap$EntryIterator.next(HashMap.java:1471)
	at java.util.HashMap$EntryIterator.next(HashMap.java:1469)
	at com.google.common.collect.AbstractMapBasedMultimap$Itr.next(AbstractMapBasedMultimap.java:1145)
	at org.osc.core.broker.job.Graph.getPredecessors(Graph.java:183)
	at org.osc.core.broker.job.TaskNode.getPredecessors(TaskNode.java:283)
	at org.osc.core.broker.job.Job.checkAllPredecessorsCompletedSuccessfully(Job.java:340)
	at org.osc.core.broker.job.Job.executeGraph(Job.java:245)
	at org.osc.core.broker.job.Job.run(Job.java:171)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:748)
### Environment
[...]

PR #588, opened 2017-10-27
JobAbort on SFC binding (ConcurrentModificationException, issue 550)

dmitryintel:
* Add unit tests that replicate this exact stack trace. 
* Tested the fix. 
* [issue 550](https://github.com/opensecuritycontroller/osc-core/issues/550)

### DIFF
--- a/osc-server/src/main/java/org/osc/core/broker/job/Graph.java
+++ b/osc-server/src/main/java/org/osc/core/broker/job/Graph.java
@@ -177,7 +177,7 @@ public class Graph<T> implements Iterable<T>, Serializable {
         return ImmutableSet.copyOf(this.edges.get(node));
     }
 
-    public Set<T> getPredecessors(T node) {
+    public synchronized Set<T> getPredecessors(T node) {
         Set<T> predecessors = new HashSet<T>();
 
         for (Map.Entry<T, T> edge : this.edges.entries()) {
@@ -211,7 +211,7 @@ public class Graph<T> implements Iterable<T>, Serializable {
     }
 
     @SuppressWarnings("unchecked")
-    public void addEdges(T src, T... dests) {
+    public synchronized void addEdges(T src, T... dests) {
         this.addNode(src);
         for (T dest : dests) {
             this.addNode(dest);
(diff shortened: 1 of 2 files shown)


osc-server/src/main/java/org/osc/core/broker/job/Graph.java
```diff
@@ -162,25 +162,37 @@ private Graph(Set<T> nodes, Multimap<T, T> edges) {
         this.edges = edges;
     }
 
-    public Set<Edge<T>> getEdges() {
+    private synchronized Set<Edge<T>> getEdges() {
         Set<Edge<T>> edgeSet = new HashSet<Edge<T>>();
-        for (Map.Entry<T, T> edge : this.edges.entries()) {
+
+        Set<Map.Entry<T, T>> copy;
```
arvindn05: you dont neeed this since the method is already synchronized

dmitryintel: I actually thought of removing synchronized.

osc-server/src/main/java/org/osc/core/broker/job/Graph.java
```diff
@@ -412,13 +406,12 @@ public boolean equals(Object obj) {
 
         @SuppressWarnings("unchecked")
         Graph<T> that = (Graph<T>) obj;
-        return new EqualsBuilder().append(this.nodes, that.nodes).append(this.edges, that.edges).isEquals();
+        return new EqualsBuilder().append(getNodes(), getNodes()).append(getEdges(), getEdges()).isEquals();
```
arvindn05: should be that.getNodes

osc-server/src/test/java/org/osc/core/broker/job/JobEngineTest.java
```diff
@@ -243,6 +246,89 @@ public void testTaskDependencyExecutionOrder() throws Exception {
         verifyJobPersistence(this.job);
     }
 
+    @Test
+    public void testConcurrentGraphModification() throws Exception {
```
arvindn05: add defect number for more detials

PR #598, opened 2017-11-01
Fix #511: Validate AZ, HostAgrregate and Host entities with create DS request

### LINKED ISSUE #511: Sync DS job (through API) "succeed" even if the host name incorrect or not exist
### Description
From the API when posting information with the selection:
    "hosts": [
      {
        "parentId": null,
        "openstackId": "not exist-comp",
        "name": "not exist-comp"
      }
    ],

### Expected Behavior
Sync job suppose to fail since there is not such a host

### Actual Behavior
Sync job pass - DAI is not created in OSC and OpenStack

### Steps to Reproduce
Through the UI create DS with one host selected.

Go to API doc
get distributed appliances in order to get VS_ID

go to virtual systems and run the get command for 
GET /api/server/v1/virtualSystems/{vsId}/deploymentSpecs

Then just change the OpenStackId and name to something invalid

Use the POST /api/server/v1/virtualSystems/{vsId}/deploymentSpecs
with the wrong host and openstack id

DS will be created and the job will pass
In the DS - when you click on edit - the hosts will be the original ones (the exist ones) but none will be selected

### Additional Information
the body for the post was:
```
            {
              "name": "default-DS",
              "projectId": "68a6ea366ffb4ea585907cd22261dbea",
              "projectName": "admin",
              "region": "RegionOne",
              "hosts": [
                    {
                        "parentId": null,
                        "openstackId": "not-exist-ocata-comp",
                        "name": "not-exist-ocata-comp"
                    }
              ],
              "availabilityZones": [],
              "hostAggregates": [],
              "managementNetworkName": "mgmt",
              "managementNetworkId": "9e8ecf25-f829-42e6-be37-67ec259a490a",
              "inspectionNetworkName": "inspection",
              "inspectionNetworkId": "06f40da8-f51d-496e-82ec-98ac7fad7b2a",
              "floatingIpPoolName" : "ext-net",
              "isShared": "True",
              "count": "1"
            }
```
### Environment
Will happen with any build

- OSC Version/Build:   Version: 1.0.0-SNAPSHOT (Build:87-gf51c7f6, 2017-10-03T07:20:46Z)
- Cloud Provider Info:
- Plugin Info:
- Other:

### Status
- [x] Reproduced
- [x] In Progress
- [x] Validated     =>    Verified in Version/Build: Version: 1.0.0-SNAPSHOT (Build:14-g7adf532, 2017-11-22T08:16:42Z)

[ServerSupportBundle.zip](https://github.com/opensecuritycontroller/osc-core/files/1354060/ServerSupportBundle.zip)

larkinscarvalho:
Fixes #511

### DIFF
--- a/osc-server/src/main/java/org/osc/core/broker/service/BaseDeploymentSpecService.java
+++ b/osc-server/src/main/java/org/osc/core/broker/service/BaseDeploymentSpecService.java
@@ -17,15 +17,21 @@
 package org.osc.core.broker.service;
 
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.persistence.EntityManager;
 
+import org.apache.commons.collections4.CollectionUtils;
 import org.osc.core.broker.model.entities.appliance.VirtualSystem;
 import org.osc.core.broker.model.entities.virtualization.openstack.AvailabilityZone;
 import org.osc.core.broker.model.entities.virtualization.openstack.DeploymentSpec;
 import org.osc.core.broker.model.entities.virtualization.openstack.Host;
 import org.osc.core.broker.model.entities.virtualization.openstack.HostAggregate;
+import org.osc.core.broker.rest.client.openstack.openstack4j.Endpoint;
+import org.osc.core.broker.rest.client.openstack.openstack4j.Openstack4JNova;
 import org.osc.core.broker.service.dto.openstack.AvailabilityZoneDto;
 import org.osc.core.broker.service.dto.openstack.DeploymentSpecDto;
 import org.osc.core.broker.service.dto.openstack.HostAggregateDto;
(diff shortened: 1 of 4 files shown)


osc-server/src/main/java/org/osc/core/broker/service/BaseDeploymentSpecService.java
```diff
@@ -78,17 +84,64 @@ protected void validate(EntityManager em, DeploymentSpecDto dto) throws Exceptio
                 throw new VmidcBrokerValidationException("Invalid count " + dto.getCount()
                 + " specified for Deployment Specification. Only valid value is 1.");
             }
+            validateSelectionCriteria(this.vs, dto);
         }
     }
 
-    protected void throwInvalidUpdateActionException(String attributeType, String dsName)
+	private void validateSelectionCriteria(VirtualSystem vs, DeploymentSpecDto dto) throws Exception {
```
arvindn05: can we use the ListAvailabilityZonesService and other services to validate instead of duplicating the code here?

Also, how about validation of other fields like tenant, floating ip, the networks etc

PR #599, opened 2017-11-02
 Fixes #596 :Check for SFC virtual system list , before binding. Need…

### LINKED ISSUE #596: When SFC is added with no virtual systems, binding SFC to a security group should fail at validation
### Description
SFC is added with no virtual systems, binding SFC to a security group should fail at validation
### Expected Behavior
SFC is added with no virtual systems, binding SFC to a security group should fail at validation
### Actual Behavior
SFC is added with no virtual systems, binding SFC to a security group is failed while binding job is executed.
### Steps to Reproduce
1. Configure SFC with no virtual systems
2. Configure a Security Group
3. Bind SFC to Security group
### Environment
- OSC Version/Build: 
osc-buikd -135
- Cloud Provider Info:
- Plugin Info:
- Other:

### Status
- [x] Reproduced
- [ ] In Progress

karimull:
… to have atleast one virtual system chained to sfc to bind


arvindn05:
@karimull please fix the build

PR #604, opened 2017-11-09
Conform svc refactor

dmitryintel:


osc-server/src/main/java/org/osc/core/broker/service/DAConformService.java
```diff
@@ -0,0 +1,198 @@
...
+import org.osc.core.broker.service.tasks.conformance.securitygroupinterface.MgrSecurityGroupInterfacesCheckMetaTask;
+import org.osc.core.broker.service.transactions.CompleteJobTransaction;
+import org.osc.core.broker.service.transactions.CompleteJobTransactionInput;
+import org.osc.core.common.job.TaskGuard;
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Reference;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Component(service = {ConformServiceApi.class, DAConformService.class})
```
arvindn05: should only expose DAConformService. ConformServiceApi should be deleted as the corresponding implementation is deleted as well

create a DAConformServiceApi, DSConformServiceAPI etc

osc-server/src/main/java/org/osc/core/broker/service/DSConformService.java
```diff
@@ -0,0 +1,200 @@
...
+
+    private BundleContext context;
+
+    @Reference
+    protected DBConnectionManager dbConnectionManager;
+
+    @Reference
+    protected TransactionalBroadcastUtil txBroadcastUtil;
+
+    private void delayedInit() {
```
arvindn05: do we need this? Refactoring of this service should have elminated the need for the delay init...

osc-server/src/main/java/org/osc/core/broker/service/DAConformService.java
dmitryintel: The DAConformService is the only one used outside the osc-server package, so adding the Api interfaces for the others will not decouple them for anyone who doesn't already know about the implementation. 

I could simply rename ConformServiceApi to DAConformServiceApi and forget about the other four. Adding all these methods that start jobs into the Api is not possible, since then osc-service-api would have to know about the Job class.

osc-server/src/main/java/org/osc/core/broker/service/AddDistributedApplianceService.java
```diff
@@ -58,7 +58,7 @@
     private DistributedApplianceDtoValidator validatorFactory;
 
     @Reference
-    private ConformService conformService;
+    private DAConformService dsConformService;
```
arvindn05: daConformService

osc-server/src/main/java/org/osc/core/broker/service/DSConformService.java
```diff
@@ -0,0 +1,200 @@
...
+    private void deactivate(BundleContext context) {
+        if (this.initDone.get()) {
+            context.ungetService(this.dsConformanceCheckMetaTaskSR);
+        }
+    }
+
+    @Reference
+    private ConformK8sDeploymentSpecMetaTask conformK8sDeploymentSpecMetaTask;
+
+    // optional+dynamic to resolve circular reference
```
arvindn05: please explain circular reference here..with this refactoring this should be resolved

osc-server/src/main/java/org/osc/core/broker/service/SGConformService.java
```diff
@@ -0,0 +1,186 @@
...
+import org.osc.core.common.job.TaskGuard;
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Reference;
+import org.osgi.service.transaction.control.ScopedWorkException;
+import org.osgi.service.transaction.control.TransactionControl;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Component(service = {SGConformService.class})
+public class SGConformService {
```
arvindn05: generally all services in OSC are service dispatchers....we should come up with a different name for the these classes SecurityGroupJobManager/Factory/?

osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/deploymentspec/DSUpdateOrDeleteMetaTask.java
```diff
@@ -93,20 +84,9 @@
     private Openstack4JNova novaApi;
     @IgnoreCompare
     private DSUpdateOrDeleteMetaTask factory;
-    @IgnoreCompare
-    private AtomicBoolean initDone = new AtomicBoolean();
 
     @Override
     protected void delayedInit() {
```
arvindn05: dont need this anymore? This code should move to the create method

osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/deploymentspec/DSUpdateOrDeleteMetaTask.java
```diff
@@ -93,20 +84,9 @@
     private Openstack4JNova novaApi;
     @IgnoreCompare
     private DSUpdateOrDeleteMetaTask factory;
```
arvindn05: do we need this factory anymore?

osc-server/src/main/java/org/osc/core/broker/service/SGConformService.java
dmitryintel: DeploymentSpecConformJobFactory, DistributedApplianceConformJobFactory. Honestly, I would love to abbreviate do DAConformJobFactory, etc.

arvindn05: we can put specific guidelines like below

https://stackoverflow.com/a/7366332/2904615

Lets not use abbreviations if possible atleast for terms internal to OSC

osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/deploymentspec/DeleteSvaServerAndDAIMetaTask.java
```diff
@@ -62,15 +53,9 @@
     private TaskGraph tg;
     @IgnoreCompare
     private DeleteSvaServerAndDAIMetaTask factory;
-    @IgnoreCompare
-    private AtomicBoolean initDone = new AtomicBoolean();
 
     @Override
     protected void delayedInit() {
```
arvindn05: same comment here....move content of this method to the create method

osc-server/src/main/java/org/osc/core/server/Server.java
```diff
@@ -128,7 +132,20 @@
     private boolean devMode = false;
 
     @Reference
-    private ConformService conformService;
+    private DistributedApplianceConformJobFactory daConformJobFactory;
+
+    @Reference(cardinality = ReferenceCardinality.OPTIONAL, policy = ReferencePolicy.DYNAMIC)
+    private volatile ComponentServiceObjects<DeploymentSpecConformJobFactory> dsConformJobFactoryCSO;
```
arvindn05: still needed?

osc-server/src/main/java/org/osc/core/broker/service/DeploymentSpecConformJobFactory.java
```diff
@@ -0,0 +1,199 @@
...
+            context.ungetService(this.dsConformanceCheckMetaTaskSR);
+        }
+    }
+
+    @Reference
+    private ConformK8sDeploymentSpecMetaTask conformK8sDeploymentSpecMetaTask;
+
+    // optional+dynamic to resolve circular reference
+    @Reference(cardinality = ReferenceCardinality.OPTIONAL, policy = ReferencePolicy.DYNAMIC)
+    private volatile ServiceReference<DSConformanceCheckMetaTask> dsConformanceCheckMetaTaskSR;
```
arvindn05: looks like something in DS has a circular dependency...lets see if we can eliminate the circular reference...

dmitryintel: I cut this one, still works. Let me commit.

PR #606, opened 2017-11-14
Fix 511: Validate AZ, HostAgrregate and Host entities with create DS …

larkinscarvalho:
…request

### DIFF
--- a/osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/deploymentspec/DSUpdateOrDeleteMetaTask.java
+++ b/osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/deploymentspec/DSUpdateOrDeleteMetaTask.java
@@ -26,6 +26,7 @@ import java.util.Set;
 
 import javax.persistence.EntityManager;
 
+import org.apache.commons.collections4.CollectionUtils;
 import org.osc.core.broker.job.Task;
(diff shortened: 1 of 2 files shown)


osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/deploymentspec/DSUpdateOrDeleteMetaTask.java
```diff
@@ -179,34 +180,45 @@ private void buildDsUpdateTaskGraph(EntityManager em) throws IOException, Except
...
 
         return hostsToDeployTo;
     }
 
     private void conformToHostsSelection(Collection<String> selectedHosts,
                                          HostAvailabilityZoneMapping hostAvailabilityZoneMap, Collection<String> osHostSet) {
+
+		List<String> invalidHosts = (List<String>) CollectionUtils.removeAll(selectedHosts, osHostSet);
+		if (!invalidHosts.isEmpty()) {
+			for (String invalidHost : invalidHosts) {
```
sudhirappaji: we use Lamda Expression instead of 'for' iteration?

sudhirappaji:
@larkinscarvalho , Can you confirm whether this as been tested with the latest Master? before I merge the changes.

larkinscarvalho:
@sudhirappaji dev validation is complete. Once you merge the changes, QA will validate the fix.

Issue #281, opened 2017-05-19
Remove job for Virtualization connector


arvindn05:
Since Virtualization connector has a job associated with it which checks SSL connectivity we should be able to force delete the VC incase the endpoint is no longer accessible.

arvindn05:
@ychoma verified the issue as part of DE4937

PR #609, opened 2017-11-17
Remove vc job

larkinscarvalho:


osc-server/src/main/java/org/osc/core/broker/service/vc/DeleteVirtualizationConnectorService.java
```diff
@@ -18,40 +18,33 @@
...
     @Override
-    public BaseJobResponse exec(BaseIdRequest request, EntityManager em) throws Exception {
+    public EmptySuccessResponse exec(BaseIdRequest request, EntityManager em) throws Exception {
         validate(em, request);
 
         OSCEntityManager<VirtualizationConnector> vcEntityMgr = new OSCEntityManager<>(VirtualizationConnector.class, em, this.txBroadcastUtil);
         VirtualizationConnector vc = vcEntityMgr.findByPrimaryKey(request.getId());
-        return new BaseJobResponse(startJob(vc));
+
+        SslCertificateAttrEntityMgr sslCertificateAttrEntityMgr = new SslCertificateAttrEntityMgr(em, this.txBroadcastUtil);
```
arvindn05: we should try write lock the vc before we delete. Otherwise if the VC is being actively used somewhere and we delete the VC underneath...its going to result in errors

osc-server/src/main/java/org/osc/core/broker/service/vc/DeleteVirtualizationConnectorService.java
```diff
@@ -65,16 +58,4 @@ void validate(EntityManager em, BaseIdRequest request) throws Exception {
         VirtualizationConnectorEntityMgr.validateCanBeDeleted(em, vc);
```
larkinscarvalho: @arvindn05 we first validate if the VC can be deleted. I think it is safe to delete VC without the write lock.

arvindn05: while we are validating it maybe safe, but when deleting it might not be. We should always write lock and hold it until the vc is safely deleted

larkinscarvalho:
@sudhirappaji PR is approved and ready to merge

sudhirappaji:
@larkinscarvalho yes will do it before your morning on Monday, I was off last Friday. Btw: I did not  get the context why job is removed.

larkinscarvalho:
@sudhirappaji please take a look at the issue https://github.com/opensecuritycontroller/osc-core/issues/281

sudhirappaji:
@larkinscarvalho , can you update the branch? and will do the merge soon after.

larkinscarvalho:
@sudhirappaji done. Thanks!

PR #615, opened 2017-11-29
Use the correct password used when modifying and saving truststore

dmitryintel:


arvindn05:
@sudhirappaji looks good to me.

PR #616, opened 2017-11-29
Fix #454: Adding VC with invalid credentials/ip does not provide warning

### LINKED ISSUE #454: Adding VC with invalid credentials/ip does not provide warning
### Description
Adding VC with invalid credentials/ip does not provide a warning allow you to ignore and continue.

This is a regression from previous behavior.

### Expected Behavior

Show a warning dialog allowing you to ignore the error and continue . The previous warning text also need to be modified to account for domain id

### Actual Behavior

shows invalid errors or just shows unauthorized 

Shows below error in case of bad VC ip. It should first complain about no connection to keystone
image

Shows unauthorized without any corrective action....or allowing you to continue
image

### Steps to Reproduce

1. Add VC with invalid ip

1. Add VC with valid ip and invalid credentials
### Environment
- OSC Version/Build: 1.0.0-SNAPSHOT-56-g3debe76
- Cloud Provider Info: Ocata openstack
- Plugin Info: n/a
- Other: n/a

### Status
- [x] Reproduced
- [x] In Progress

larkinscarvalho:
Adding VC with invalid credentials/ip does not provide warning

## Description
Adding VC with invalid credentials/ip does not provide a warning allow you to ignore and continue.

## Motivation and Context
https://github.com/opensecuritycontroller/osc-core/issues/454

## How Has This Been Tested?
Manually verified. Unit tests are passing.

### DIFF
--- a/osc-server/src/main/java/org/osc/core/broker/util/VirtualizationConnectorUtil.java
+++ b/osc-server/src/main/java/org/osc/core/broker/util/VirtualizationConnectorUtil.java
@@ -19,10 +19,13 @@ package org.osc.core.broker.util;
 import static org.osc.core.common.virtualization.VirtualizationConnectorProperties.ATTRIBUTE_KEY_HTTPS;
 
 import java.io.IOException;
+import java.net.ConnectException;
 import java.util.ArrayList;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicBoolean;
 
+import org.openstack4j.api.exceptions.AuthenticationException;
+import org.openstack4j.api.exceptions.ConnectionException;
 import org.osc.core.broker.model.entities.virtualization.VirtualizationConnector;
 import org.osc.core.broker.model.plugin.ApiFactoryService;
 import org.osc.core.broker.rest.client.k8s.KubernetesClient;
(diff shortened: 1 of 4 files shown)


larkinscarvalho:
@sudhirappaji Issue was introduced caused by regression. And yes checkOpenstackConnection validates connections with keystone.

osc-server/src/main/java/org/osc/core/broker/util/VirtualizationConnectorUtil.java
```diff
@@ -194,7 +197,11 @@ private void delayedInit() {
                 this.keystoneApi.listProjects();
 
             } catch (Exception exception) {
-                errorTypeException = new ErrorTypeException(exception, ErrorType.PROVIDER_EXCEPTION);
+				if (exception instanceof ConnectionException || exception instanceof AuthenticationException) {
```
arvindn05: why is authentication exception a connect exception? we should provide the correct exceptions and messages for connection vs auth exceptions.

osc-ui/src/main/java/org/osc/core/broker/view/vc/BaseVCWindow.java
```diff
@@ -303,13 +303,20 @@ protected void handleException(final Exception originalException) {
...
             // TODO this exception leaks large amounts of implementation detail out of the API
-            if (errorType == ErrorType.PROVIDER_EXCEPTION) {
-                if (RestClientException.isConnectException(exception) && isOpenstack()) {
-                    // Keystone Connect Exception
-                    contentText = VmidcMessages.getString(VmidcMessages_.VC_CONFIRM_IP, KEYSTONE_CAPTION);
-                }
-            } else if (errorType == ErrorType.CONTROLLER_EXCEPTION) {
-                if (RestClientException.isCredentialError(exception)) {
+			if (errorType == ErrorType.PROVIDER_EXCEPTION) {
+				if (RestClientException.isConnectException(exception) && isOpenstack()) {
```
arvindn05: since we are checking isOpenstack for all the conditions, can we re-structure this if-else ladder to be more consice and clear?

if(isOpenstack()) {
switchcase or a if-else ladder
}

larkinscarvalho:
@arvindn05 Yes. I looked at it and makes sense. I will update the PR.

sudhirappaji:
@larkinscarvalho , could you please update code based on Arvind's feedback?

Issue #418, opened 2017-08-09
Selection Type for Security Group by Type displays Vm - it should be displayed as VM (as VM is an abbreviation for Virtual Machines)


mrajora-1:
### Steps to Reproduce
On OSC/Setup/Virtualization Connector/, add  a Security Group. 
Selection type column displays Vm. 
It should be displayed as VM as VM is an abbreviation for Virtual Machines. 
Earlier it was 'VM'. Dont know how it got changed to Vm. 
We should change it back to 'VM'

Screen shot should suffice for this bug. 

### Additional Information
DE4038 in Rally 

### Environment
- OSC Version/Build: 0.6 (Build:29-g0adc599, 2017-05-09T01:59:44Z)
- Cloud Provider Info:
- Plugin Info:
- Other:

### Status
- [ ] Reproduced
- [ ] In Progress
- [ ] Validated     =>    Verified in Version/Build: 

![sg-displaying-vm-instead-of-vm](https://user-images.githubusercontent.com/28600997/29148374-b38b866c-7d21-11e7-81e1-206f74a9e2f9.JPG)

pravinrajr9:
This issue can no more reproducible in latest OSC version
Tested with
[sg create](https://user-images.githubusercontent.com/30862177/34215389-3778c144-e5cb-11e7-872b-2bb4258fa789.png)
 Build: Build:171-g9da3d6a, 2017-12-18T06:25:16Z

PR #624, opened 2017-12-17
Fix ui defects

pravinrajr9:
Fixing of OSC UI defect #418

## Description
Fixed the OSC UI defect for displaying selection type option parameters correctly,

## Motivation and Context

fixes issue https://github.com/opensecuritycontroller/osc-core/issues/418

## How Has This Been Tested?
Tested manually

## Types of change

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [X] My code follows the [Code style guidelines](https://github.com/opensecuritycontroller/community/blob/master/development/eclipse.md) of this project
- [X] Unit test coverage percentage is maintained(increased/remains the same but not decreased)


arvindn05:
@pravinrajr9 not sure how this fixes the original issue. The original issue is about the text Vm -> VM

pravinrajr9:
@arvindn05 I added this parameter assuming it could reflect the changes immediately since everything was fine in the code. But it worked fine for older version of osc-ui, where I was able to reproduce the issue.  I am not able to reproduce this defect in latest versions of osc-ui. We can close this bug.  It may be a good practice to add this parameter since we want listener to get called immediately for workload selection type, this pull request may be irrelevant to this defect as it can not be reproduced in newer osc versions.

arvindn05:
`https://vaadin.com/forum/#!/thread/2983812/3070078`
@pravinrajr9 Immediate has the cost of as soon as the field changes value, it is sent to the server. 

I think here the combo box results in openstack calls which are expensive. Not having it immediate is better in this case as we make the call only after the user moves away from the combobox.

arvindn05:
@pravinrajr9 Please close the pull request as it seems this is not needed for the bug fix. You can mark the bug as non-issue and post a screenshot there as collateral.

Issue #627, opened 2018-01-06
Removing VMs from bound Security Group results in referrential integrity error


dmitryintel:
### Description
Deleting VM members from SG while it is bound sometimes fails with referential integrity exceptions.

### Expected Behavior
Should be able to edit SG members while bound.

### Actual Behavior

Removing the original VM fails. 

### Steps to Reproduce
Openstack (mitaka) with two VMS is required. Reproduced on Newton with NSC.

- Create a Security Group with one Virtual Machine. 
- Bind.
- Edit the a second VM while bound. 
- Remove the original VM while bound.
  (or try removing both)

### Additional Information
Stack trace:

2018-01-05 18:48:57,205 [INFO | isc-task-pool-5| o.o.c.b.s.t.c.o.s.SecurityGroupMemberDeleteTask]: Deleting Security Group member from SGNew01
2018-01-05 18:48:57,274 [WARN | isc-task-pool-5| o.o.c.b.j.TaskNode]: Task TaskNode [task=[null], state=RUNNING, status=PASSED, guard=ALL_PREDECESSORS_SUCCEEDED] Failed
org.osgi.service.transaction.control.TransactionRolledBackException: The transaction rolled back due to a failure
        at org.apache.aries.tx.control.service.common.impl.AbstractTransactionControlImpl$TransactionBuilderImpl.doWork(AbstractTransactionControlImpl.java:208)
        at org.apache.aries.tx.control.service.common.impl.AbstractTransactionControlImpl$TransactionBuilderImpl.required(AbstractTransactionControlImpl.java:84)
        at org.apache.aries.tx.control.service.common.impl.AbstractTransactionControlImpl.required(AbstractTransactionControlImpl.java:263)
        at org.osc.core.broker.service.tasks.TransactionalTask.execute(TransactionalTask.java:44)
        at org.osc.core.broker.job.TaskNode.run(TaskNode.java:317)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:748)
Caused by: org.osgi.service.transaction.control.TransactionException: An error occurred when committing the connection
[...]

PR #630, opened 2018-01-09
Fixing bug #627: can delete bound SGs now.

dmitryintel:
Deleting bound ("binded") security groups or removing members from them -- whithout unbinding first -- used to result in referential integrity exceptions. This was reported as [issue #627](https://github.com/opensecuritycontroller/osc-core/issues/627)

## Description

Changed SecurityGroupMemberDeleteTask to dissociate security group members from ports, networks, etc. Carefully severe the foreign key ties to prevent ReferrentialIntegrityException.

## Motivation and Context

Fixes [issue #627](https://github.com/opensecuritycontroller/osc-core/issues/627).
## How Has This Been Tested?

Tested with OSC PAN plugin, NSC plugin (custom version which supports port groups), Panorama 8.0 and Mitaka Openstack.

- Checked that SG with several Virtual Machine (VM) members can be bound and withstand VM being removed. .
- Checked that SG with several VM members can be bound and deleted.
- Checked that SG with a single Network or Subnet member can be bound and deleted.
Jobs for all these changes used to fail before this fix.

## Types of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] My code follows the [Code style guidelines](https://github.com/opensecuritycontroller/community/blob/master/development/eclipse.md) of this project
- [x] My unit test follow the [Unit test guidelines](https://github.com/opensecuritycontroller/community/blob/master/development/unit_test_guidelines.md)
- [x] Unit test coverage percentage is maintained(increased/remains the same but not decreased)

### DIFF
--- a/osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/securitygroup/SecurityGroupMemberDeleteTask.java
+++ b/osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/securitygroup/SecurityGroupMemberDeleteTask.java
@@ -16,6 +16,7 @@
  *******************************************************************************/
 package org.osc.core.broker.service.tasks.conformance.openstack.securitygroup;
 
+import java.util.HashSet;
 import java.util.Set;
 
 import javax.persistence.EntityManager;
@@ -33,9 +34,9 @@ import org.osc.core.broker.model.entities.virtualization.openstack.VM;
 import org.osc.core.broker.model.entities.virtualization.openstack.VMPort;
 import org.osc.core.broker.service.persistence.OSCEntityManager;
 import org.osc.core.broker.service.tasks.TransactionalTask;
-import org.slf4j.LoggerFactory;
 import org.osgi.service.component.annotations.Component;
 import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 @Component(service=SecurityGroupMemberDeleteTask.class)
 public class SecurityGroupMemberDeleteTask extends TransactionalTask {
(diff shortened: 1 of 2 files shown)


osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/securitygroup/SecurityGroupMemberDeleteTask.java
```diff
@@ -160,4 +141,66 @@ public String getName() {
...
+    }
+
+    private void deleteFromDB(EntityManager em, Network network) {
+        Set<VMPort> ports = new HashSet<>(network.getPorts());
+        for (VMPort vmPort : ports) {
+            // If the VM was created on behalf of this port, delete it.
+            VM vm = vmPort.getVm();
+            deleteFromDB(em, vmPort);
+            if (vm != null) {
+                vm.removePort(vmPort);
```
arvindn05: isnt this done twice once in the "deleteFromDB(em, vmPort)" call? i think this would be a noop.

osc-server/src/main/java/org/osc/core/broker/service/tasks/conformance/openstack/securitygroup/SecurityGroupMemberDeleteTask.java
```diff
@@ -160,4 +141,64 @@ public String getName() {
...
+     * This method is only called when all the ports are gone.
+     * @param em
+     * @param vm
+     */
+    private void deleteFromDB(EntityManager em, VM vm) {
+        this.log.info("Removing vm " + vm);
+        OSCEntityManager.delete(em, vm, this.txBroadcastUtil);
+    }
+
+    private void deleteFromDB(EntityManager em, Network network) {
```
arvindn05: doesnt look like any difference between this and deleteFromDB(EntityManager em, Network network).

Can we just create a method deleteFromDB(EntityManager em, Set<VMPort> ports)?

PR #639, opened 2018-01-25
Build automation to generate OSC docker image

pravinrajr9:
The build.xml is now updated to support OSC containerization, The user can get the OSC docker image in tar format.  
## Description

The user need to run below command to get OSC as a docker image in tar format
go to osc-core
`ant dockerimage`
The docker image will  be generated in Build-xxx directory in a tar format
## Motivation and Context

To support OSC Containerization
## How Has This Been Tested?

Tested Manually
## Types of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [X] My code follows the [Code style guidelines](https://github.com/opensecuritycontroller/community/blob/master/development/eclipse.md) of this project

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -177,7 +177,7 @@
     <target name="createDockerImage" description="create docker image">
         <exec executable="docker" resultproperty="tar.result">
             <arg value="pull" />
-            <arg value="centos:7" />
+            <arg value="centos:latest" />
         </exec>
         <exec executable="docker" resultproperty="tar.result" dir="docker/">
             <arg value="build" />
--- a/build.xml
+++ b/build.xml
@@ -19,7 +19,7 @@
 
 	<property name="image-format" value="vmdk_only" />
 	<target name="ovf" depends="build,ovfOnly,ovfCopy" description="build all and generate vmidc ovf" />
-
+        <target name="dockerimage" depends="build,dockerImageBuild,clean" description="build all and generate docker image in tar" />
 	<target name="build" depends="deleteJavaJre">
 		<echo>Ant version: ${ant.version}</echo>
 		<echo>Ant Java/JVM version: ${ant.java.version}</echo>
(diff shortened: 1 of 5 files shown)


pravinrajr9:
@sudhirappaji @arvindn05  please review

build.xml
```diff
@@ -19,7 +19,7 @@
 
 	<property name="image-format" value="vmdk_only" />
 	<target name="ovf" depends="build,ovfOnly,ovfCopy" description="build all and generate vmidc ovf" />
-
+        <target name="dockerimage" depends="build,dockerImageBuild,clean" description="build all and generate docker image in tar" />
```
sudhirappaji: why can't we separate out the build.xml for the container?

build.xml
```diff
@@ -132,17 +132,76 @@
...
+        <!-- Version examples: 1.0.0-SNAPSHOT, 1.0.0, etc -->
+        <exec outputproperty="releaseVersion" dir="." executable="sh">
+            <arg value="-c" />
+            <arg value="git describe --abbrev=0 | tr -d v" />
+        </exec>
+        <echo>Full Version: ${fullVersion}</echo>
+        <echo>Version: ${releaseVersion}</echo>
+    </target>
+
+    <target name="dockerImageBuild" depends="prodVersion,createBuildDir,copyAndUnzipServerUpgradeBundle,copyServerUpgradeBundle,createDockerImage" description="Copy result files to build dir">
```
sudhirappaji: I see duplication of logic and it is better to keep separate xml for docker image automation process.

build.xml
```diff
@@ -132,17 +132,76 @@
...
+        </zip>
+    </target>
+
+    <target name="checkServerUpgradeBundleExist">
+        <available file="docker/serverUpgradeBundle*.zip" property="isServerUpgradeBundleExist"/>
+    </target>
+
+    <target name="deleteOldServerUpgradeBundle" depends="checkServerUpgradeBundleExist" if="isServerUpgradeBundleExist">
+        <delete>
+            <fileset dir="docker/" includes="serverUpgradeBundle-*.zip" />
```
sudhirappaji: Can we have configuration items defined, where we can specify the folder names? rather than hard-coding it. Later on when we start to use Kolla Container, it would be easy to change in configuration file, rather than in the build file. also it is easier to maintain.

build.xml
```diff
@@ -132,17 +132,76 @@
...
+
+    <target name="clean" >
+        <delete file="docker/serverUpgradeBundle-${fullVersion}.zip"/>
+        <delete dir="docker/opt" />
+    </target>
+
+    <target name="createDockerImage" description="create docker image">
+        <exec executable="docker" resultproperty="tar.result">
+            <arg value="pull" />
+            <arg value="centos:7" />
```
sudhirappaji: why is the version specified? I presume we use latest available centos from centos repository.

docker/Dockerfile-osc-base
```diff
@@ -0,0 +1,25 @@
+FROM centos:7
+
+MAINTAINER Arvind Nadendla <arvindn05@gmail.com>
+
+ENV http_proxy http://proxy-chain.intel.com:911
```
sudhirappaji: proxy cannot be hardcoded, rather it should be placed in configuration. or we add a note, so that user can change to no-proxy or relevant proxy depending on the need.

docker/Dockerfile-osc-base
```diff
@@ -0,0 +1,25 @@
...
+ENV https_proxy https://proxy-chain.intel.com:912
+
+RUN yum update -y && \
+yum install -y wget && \
+wget --no-check-certificate --no-cookies --header "Cookie: oraclelicense=accept-securebackup-cookie" http://download.oracle.com/otn-pub/java/jdk/8u161-b12/0da788060d494f5095bf8624735fa2f1/jre-8u161-linux-x64.rpm && \
+yum localinstall -y jre-8u161-linux-x64.rpm && \
```
sudhirappaji: do we need to hardcode the jre version? can this be part of the configuration?

build.xml
pravinrajr9: Done, now docker pulls latest centos image from docker repository

docker/Dockerfile-osc-base
pravinrajr9: Done, I created a new proxy.sh file, where user can configure proxy settings

docker/Dockerfile-osc-base
pravinrajr9: I need to figure out better approach. There is no installer command for oracle java jre in CentOS.

build.xml
pravinrajr9: @sudhirappaji  I followed legacy approach for building the image within osc-core.  Since osc-core can contain only one build.xml file, it may also confuse user if we have multiple build.xml files in same module. Please re-consider and let me know for further changes.

sudhirappaji: combining both methods into one does not help... it is better to separate. lets get opinion from Arvind and Emanoel, how they look at this?

arvindn05: given that ovf build xml's will be removed shortly, i would go with whatever is quickest to implement. To sudhir's poiint, Please keep an eye out so that we are able to strip out the ovf build portions easily without impacting the new docker portions use sections of the file etc

pravinrajr9:
Adding shutdown operation when new OSC internal keypair is replaced, @sudhirappaji please review.

PR #644, opened 2018-02-08
Adding support for docker build in travisCI

pravinrajr9:
## Description

Adding support for docker build in travisCI
## Motivation and Context

To check docker image build automation through Travis CI
## How Has This Been Tested?

Manually testet
## Types of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [ ] My code follows the [Code style guidelines](https://github.com/opensecuritycontroller/community/blob/master/development/eclipse.md) of this project


docker/Dockerfile-osc-base
```diff
@@ -2,24 +2,42 @@ FROM centos:latest
...
-COPY proxy.sh /etc/profile.d/
-
-RUN source /etc/profile.d/proxy.sh && \
-yum update -y && \
+RUN yum update -y && \
 yum install -y wget && \
 wget --no-check-certificate --no-cookies --header "Cookie: oraclelicense=accept-securebackup-cookie" http://download.oracle.com/otn-pub/java/jdk/8u162-b12/0da788060d494f5095bf8624735fa2f1/jre-8u162-linux-x64.rpm && \
 yum localinstall -y jre-8u162-linux-x64.rpm && \
 rm -rf jre-8u162-linux-x64.rpm && \
+wget --no-check-certificate --no-cookies --header "Cookie: oraclelicense=accept-securebackup-cookie" http://download.oracle.com/otn-pub/java/jdk/8u162-b12/0da788060d494f5095bf8624735fa2f1/jdk-8u162-linux-x64.rpm && \
```
arvindn05: oracle JDK cannot be distributed as a docker image. Please use JRE only.

pravinrajr9: @arvindn05 maven command do not build OSC if we use jre, I think we can remove the jdk once the OSC is built.

arvindn05: this puts JDK into the docker image file. we cannot have that. Compiling of OSC happens outside of docker so we should not required JDK within the docker.

arvindn05: ok. can we use multi stage build for building OSC? 

If you are not using multi stage build, we dont need to build osc in the docker image.

also travis is failing.

pravinrajr9: Now I have removed the OSC build part in docker image.  
I thought of making multi stage build. Problems with multistage build are travis generates the log exceeding 4Mb which causes travis build to fail, its a limitation in travis, need to make extra work around for it and also the kolla uses docker version  that does not support multi stage build.  The reason for travis build failure was due to exceeding log output since OSC is built two times once for jacoco analysis and another inside docker image.

.travis.yml
```diff
@@ -47,5 +56,10 @@ script:
       else
         mvn org.jacoco:jacoco-maven-plugin:prepare-agent install org.jacoco:jacoco-maven-plugin:report sonar:sonar;
       fi
+
+    - export http_proxy=""
+    - export https_proxy=""
+    - ant dockerimage
```
arvindn05: looks like we are building OSC twice. 

We should just be able to build OSC once(with `mvn org.jacoco:jacoco-maven-plugin:prepare-agent install org.jacoco:jacoco-maven-plugin:report sonar:sonar`) and add the output binary to the docker image.

docker/Dockerfile-osc-base
```diff
@@ -2,21 +2,17 @@ FROM centos:latest
...
-yum update -y && \
+RUN yum update -y && \
 yum install -y wget && \
 wget --no-check-certificate --no-cookies --header "Cookie: oraclelicense=accept-securebackup-cookie" http://download.oracle.com/otn-pub/java/jdk/8u162-b12/0da788060d494f5095bf8624735fa2f1/jre-8u162-linux-x64.rpm && \
 yum localinstall -y jre-8u162-linux-x64.rpm && \
 rm -rf jre-8u162-linux-x64.rpm && \
 yum install -y iproute && \
 yum install -y graphviz && \
-rm -rf /var/cache/yum
+rm -rf /var/cache/yum
```
arvindn05: extra space

.travis.yml
```diff
@@ -47,5 +52,12 @@ script:
       else
         mvn org.jacoco:jacoco-maven-plugin:prepare-agent install org.jacoco:jacoco-maven-plugin:report sonar:sonar;
       fi
+
+    - cp target/serverUpgradeBundle.zip docker/
+    - cd docker
+    - unzip serverUpgradeBundle.zip
+    - docker build -f Dockerfile-osc-base . -t osc-base-image:osc-base
```
arvindn05: can we just execute `createDockerImage` ant target here? we are pretty much doing the same thing here manually

arvindn05:
@pravinrajr9 Please make sure the changes above can be used by jenkins to generate container images as well. Is this being validated on the BLR jenkins or the SC Jenkins?

pravinrajr9:
Hi @arvindn05, this has not been validated through Jenkins. Will validate it

Issue #632, opened 2018-01-09
Security Group VM members unable to sync with OpenStack


pravinrajr9:
### Description
Security Group VM Member is not able to sync with OpenStack.
When a Security group is created with one or more VMs and when the VM is deleted through OpenStack Horizon, this should decrement the No of VM members count in Security group. The sync functionality passes and also fails. This Happens only with OpenStack Pike Version.

### Expected Behavior
The Security Group VM member should be deleted.
In the code SecurityGroupMemberDeleteTask function should be called
![image](https://user-images.githubusercontent.com/30862177/34730184-ac817b34-f584-11e7-912f-87ec4002d91e.png)

### Actual Behavior
The Security Group VM member port is deleted, VM member is not deleted.
In Code SecurityGroupMemberVmUpdateTask function is called.
![image](https://user-images.githubusercontent.com/30862177/34730245-d60f7596-f584-11e7-86f4-ef30f1f8cd24.png)

### Steps to Reproduce
1. Create VC with NSC as pluging
2. Create SG with any no of VMs
3. Go to OpenStack Horizon and delete the same VM
4. Check the Security group VM member count
To reproduce the bug, create SG with more than one VM, and repeat step 3 multiple times until the bug is found

### Additional Information
Workaround manually click sync button in Security group panel, the VM member count is updated.

### Environment
- OSC Version/Build: Version: 0.8.0 (Build:1-g62f2b90, 2017-12-01T19:51:38Z)
- Cloud Provider Info:
- Plugin Info:
- Other:  OpenStack Pike

### Status
- [ ] Reproduced
- [ ] In Progress
- [ ] Validated     =>    Verified in Version/Build: 

[ServerSupportBundle (10).zip](https://github.com/opensecuritycontroller/osc-core/files/1615775/ServerSupportBundle.10.zip)

sudhirappaji:
@bmmukund , please investigate and if needed, discussed with Dmitry.

bmmukund:
I am looking into this issue.

dmitryintel:
A sure way to reproduce is to create 4 or 6 servers, add them to the security group, and then delete them from command line on openstack in quick succession: 
```
(openstack) server delete vm1
(openstack) server delete vm2
(openstack) server delete vm3
(openstack) server delete vm4
```

dmitryintel:
[...]

PR #645, opened 2018-03-02
Add job buffering for sgsync jobs on RabbitMq Port messages

dmitryintel:
## Description
Upon RabbitMQ messsage about changing ports in openstack, a security group which protects those ports should be synchronized. If we delete multiple protected vms from openstack in quick succession, the SG in the OSC database does not reflect the correct number of the remaining vms.

This was reported in [issue 632](https://github.com/opensecuritycontroller/osc-core/issues/632).

## Motivation and Context

The Security Group membership in OSC should reflect the correct state of the actual openstack. When we delete a single vm from openstack, a RabbitMQ message is sent to OSC and it kicks off a SG Sync job. 

However, if we delete multiple ports, multiple sync jobs are attempted and the way the code is written, later submission attempts fail. Then the last successful sync job may have run without taking into account some VM deletions. This results in the Security Group object not knowing about some of its members missing.

An obvious workaround is to just press the Sync button when the dust settles, but we can buffer sync requests as shown in this PR.

## How Has This Been Tested?
Tested with ISM and NSC plugins. 

- Created a SG with several (5 to 7) vm members. 
- Binded. 
- Deleted several vms, one by one from openstack command line. Command line  allows us to do it very quickly.

## Types of change

`OsPortNotificationListener` buffers multiple sync requests. All we have to do is make sure a good sync job is started after changes have stopped coming in.

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [ ] My code follows the [Code style guidelines](https://github.com/opensecuritycontroller/community/blob/master/development/eclipse.md) of this project

### DIFF
--- a/osc-server/src/main/java/org/osc/core/broker/rest/client/openstack/vmidc/notification/listener/OsPortNotificationListener.java
+++ b/osc-server/src/main/java/org/osc/core/broker/rest/client/openstack/vmidc/notification/listener/OsPortNotificationListener.java
@@ -16,6 +16,8 @@
  *******************************************************************************/
 package org.osc.core.broker.rest.client.openstack.vmidc.notification.listener;
 
+import static org.osc.core.broker.rest.client.openstack.vmidc.notification.listener.OsSyncJobSubmitter.triggerSGSync;
+
 import java.util.List;
 
 import javax.persistence.EntityManager;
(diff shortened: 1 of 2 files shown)


arvindn05:
as discussed, the solution for the problem would be to enable granular sync at the member level instead of SG level. Please add the comments discussed and close this pull request.

dmitryintel:
Closing the PR for now. Bug to be addressed as a larger story at some later point. See issue #

dmitryintel:
Closing the PR for now. Bug to be addressed as a larger story at some later point. See issue [#632](https://github.com/opensecuritycontroller/osc-core/issues/632).

dmitryintel:
Done. https://github.com/opensecuritycontroller/osc-core/pull/645
Commented on issue. Issue itself not closed.

Dmitry.

